PR #3, opened 2020-09-22
variable names detection post-solving. some refactoring

pchtsp:


pchtsp (commented):
una pregunta que tengo es... ¿Deberíamos hacer un paquete con el cliente python de cornflow fuera del proyecto principal? Así podríamos usarlo en las aplicaciones que hacen uso de cornflow.
Cornflow podría luego tener el cliente como dependencia. Básicamente quitaríamos este fichero y lo pasaríamos a un paquete de python que desplegaríamos en pypi.

airflow_config/dags/cornflow_api.py
```diff
@@ -1,6 +1,6 @@
 import requests
```
ggsdc, line 1: This file should be deleted because now it is here: https://github.com/baobabsoluciones/cornflow-client

airflow_config/dags/optim_dag.py
```diff
@@ -1,7 +1,7 @@
 from airflow import DAG
```
ggsdc, line 1: These changes are mine, check them @pchtsp

examples/basic_functions.py
```diff
@@ -7,6 +7,7 @@
 config = dict(email=email, pwd=pwd, name=name)
 
 client = CornFlow(url="http://127.0.0.1:5000")
```
ggsdc, line 9: I would not have this url appear her, although it has appeared in the client repository so we should ask if David can change it

examples/pulp_varnames.py
```diff
@@ -0,0 +1,138 @@
+from cornflow_client import CornFlow, group_variables_by_name
```
ggsdc, line 1: Maybe we should move this example to the new repository

tests/test_varnames.py
```diff
@@ -0,0 +1,103 @@
+import unittest
```
ggsdc, line 1: This file is in the new repository, so it could be deleted as well @pchtsp

examples/pulp_varnames.py
pchtsp: I think maybe we should have a `requirements-dev.txt` for cornflow that includes `cornflow_client` so it can be tested. This way we can keep the tests in cornflow.

airflow_config/dags/optim_dag.py
pchtsp: For the `solver` functionality of cornflow (`optim_dag.py`) I think we should just leave one task, because there is nothing to do before or after solving and it increases time.
For other, more complicated, dags (specific to problems) this makes sense since the machines with expensive solvers may be a bottleneck and the pre-processing and post-processing could be done somewhere else without license. We will have to manage nevertheless the data between the machines...

airflow_config/dags/cornflow_api.py
pchtsp: ok, I'll make a new commit with these changes and some of mine.

airflow_config/dags/optim_dag.py
ggsdc: Ok, we can simplify this one. Undo my changes and delete the preceding task and do something like that for future dags

examples/pulp_varnames.py
ggsdc: Ok, as it can be useful on development but it should not be needed on deployment

PR #6, opened 2020-10-28
Feature/docstrings generalization

ggsdc:
I open this PR so it is ready to be merged when the last commit is done, but it is not ready yet.

The changes are:

* Docstrings in all the code.
* Generalization of models with abstract models so if we create new one they will always have certain columns (created_at, modified_at and user_id) Should we add a deleted_at field?
* Generalization of views with the class BaseResource (it already exists but it does nothing).


pchtsp:
remember to rename the `flaskr` directory to `cornflow`.

PR #7, opened 2020-10-29
Example of resource generalization

ggsdc:
This is an example of how endpoint generalization can be done and have a meta resource from which all endpoints depend.
In this case is to perform a simple get all info from the logged in user, but this could be extended to any endpoint.

This is because most endpoints should only need:
* The base model.
* The query to the database (either user or a particular object)
* The schema to validate the data.

If you need more examples to decide if you want to go with this route I can generalize more this endpoint or another so we can see that it is compatible with any endpoint

Tell me what you think

### DIFF
--- a/flaskr/endpoints/instance.py
+++ b/flaskr/endpoints/instance.py
@@ -8,6 +8,7 @@ from flask import request
 from flask_restful import Resource
 
 # Import from internal modules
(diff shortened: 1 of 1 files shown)


ggsdc:
You should check the other PR as the meta_resource file is in there as I added it by mistake in an earlier commit.

PR #10, opened 2020-12-09
Feature/unit testing

ggsdc:
Example of how to perform unit testing in flask.

Current coverage is (this is only testing log in and sing up and running the app):

image


flaskr/app.py
```diff
@@ -36,7 +39,7 @@ def create_app(env_name):
     return app
 
 if __name__ == '__main__':
-    # env_name = os.getenv('FLASK_ENV')
-    env_name = 'development'
-    app = create_app(env_name)
+    environment_name = os.getenv('FLASK_ENV')
```
pchtsp, line 42: no default environment ?

requirements.txt
```diff
@@ -1,31 +1,51 @@
...
+Pygments==2.7.2
 PyJWT==1.7.1
+pyparsing==2.4.7
 python-dateutil==2.8.1
 python-editor==1.0.4
 pytz==2020.1
 requests==2.24.0
 six==1.14.0
+snowballstemmer==2.0.0
+Sphinx==3.2.1
```
pchtsp, line 40: sphinx should not be in `requirements.txt` but a `requirements-dev.txt`. We should check the other dependencies. Everything that is used to test should also be in `-dev.txt`.

README.rst
```diff
@@ -316,3 +316,24 @@ destroy the machine with::
 
 cornflow app  "http://vagrantfileIP:5000"
 airflow GUI  "http://vagrantfileIP:8080"
+
+Test conrflow
+~~~~~~~~~~~~~~~~~~
+
+To test conrflow first you will have to create a new database::
+
+    sudo su - postgres
```
pchtsp, line 325: couldn't we do this inside the testing command? so the user only runs one comment in order to test everything?

requirements.txt
```diff
@@ -1,31 +1,51 @@
...
 alembic==1.4.2
 aniso8601==8.0.0
+Babel==2.8.0
 bcrypt==3.1.7
 certifi==2020.6.20
 cffi==1.14.0
 chardet==3.0.4
 click==7.1.2
+colorama==0.4.4
+coverage==5.3
```
pchtsp, line 11: this should also go in requirements-dev.txt I think.

pchtsp (changes requested):

flaskr/app.py
ggsdc: I have changed so development is the default one

README.rst
ggsdc: I could try to create an executable file that creates the database, runs the tests and then deletes the database, but said executable should have a linux version and a win version

requirements.txt
ggsdc: I will fix the requirements files and try to have them correctly separated

pchtsp:
pero no lo squasheaste??

pchtsp:
ah sí perdona, no vi bien. Gracias!

PR #11, opened 2020-12-09
Feature/docs endpoint generalization

ggsdc:
All the changes from both PR, documentation creation and endpoint generalization and general cornflow tidy up


ggsdc:
All I could test, as I do not have an airflow web server seems to work but more extensive testing should be done once we upgrade to airflow 2.0.

I will merge this PR and continue working on the documentation and unit test coverage that right now sits at 78%.

PR #16, opened 2021-01-15
Feature/unit testing

ggsdc:
PR with unit testing improved and extended. Current test coverage is around 88%

Important changes: 

- The creation of an execution url now expects a query parameter on the url if it is going to be solved or not. By defect it is not solved (this can be changed), the url now should be: /execution/?run=no or /execution/?run=yes.
- A new column has been added to the execution model (dag_run_id) to store the id generated by airflow on running.
- The first migration file has been added to the repo, this means that all databases should be dropped and recreated (py manage.py db upgrade) in order to have all of them synchronized


cornflow/tests/custom_test_case.py
```diff
@@ -40,50 +42,108 @@ def create_new_row(self, file):
...
+                self.payload[key] = value
+
         response = self.client.post(self.url, data=json.dumps(self.payload), follow_redirects=True,
                                     headers={"Content-Type": "application/json",
                                              "Authorization": 'Bearer ' + self.token})
 
+        if response.status_code == 400:
+            print(response.json, self.url)
+            print(self.payload)
+            print(self.token)
```
pchtsp, line 56: take out prints

cornflow/tests/custom_test_case.py
```diff
@@ -31,6 +31,8 @@ def setUp(self):
         self.url = None
         self.model = None
         self.payload = None
```
pchtsp, line 33: payload is not needed as a property. It's always read and used in the same function.

cornflow/tests/custom_test_case.py
```diff
@@ -40,50 +42,108 @@ def create_new_row(self, file):
...
 
         rows = self.client.get(self.url, follow_redirects=True,
                                headers={"Content-Type": "application/json", "Authorization": 'Bearer ' + self.token})
 
-        for i in range(len(self.payload)):
+        self.assertEqual(len(rows.json), len(files))
+
+        for i in range(len(data)):
             self.assertEqual(rows.json[i]['id'], codes[i])
-            for key in self.payload[i]:
```
pchtsp, line 73: why not just do `self.assertEqual(rows.json, payload)`?

cornflow/tests/custom_test_case.py
```diff
@@ -40,50 +42,108 @@ def create_new_row(self, file):
...
 
-    def get_one_row(self):
-        pass
+        row = self.client.get(self.url, follow_redirects=True,
+                              headers={"Content-Type": "application/json", "Authorization": 'Bearer ' + self.token})
+
+        self.assertEqual(200, row.status_code)
+        self.assertEqual(row.json['id'], self.id)
+        for key in self.payload:
+            self.assertEqual(row.json[key], self.payload[key])
```
pchtsp, line 101: same here: why not `self.assertEqual(row.json, payload)`?

cornflow/tests/unit/test_instances.py
```diff
@@ -0,0 +1,79 @@
...
+    def test_new_instance(self):
+        self.create_new_row(INSTANCE_PATH)
+
+    def test_get_instances(self):
+        self.get_rows(INSTANCES_LIST)
+
+    def test_get_no_instances(self):
+        self.get_no_rows()
+
+
```
pchtsp, line 24: missing (maybe add it as a TODO?);
* create an incorrect instance and validate the response. 
* create an incomplete instance.
* get a non-existent instance.
* all the above, for executions.

cornflow/endpoints/execution.py
```diff
@@ -63,8 +63,9 @@ def post(self):
         self.user_id, self.admin, self.super_admin = Auth.return_user_info(request)
         result = self.post_list(request)
 
-        # TODO: check result[1] before continuing! if not 20X, return.
-        if result[1] >= 300:
+        not_run = request.args.get('run', 'no') == 'no'
```
pchtsp, line 66: why a string and not a boolean? I think it should be a 1, 0 in the request args. And called `/execution/?run=0 or /execution/?run=1` with default=1.

pchtsp (changes requested):
Buenas! gracias por el PR. Algunos cambios y preguntas abajo.

cornflow/endpoints/execution.py
ggsdc: It can be a number in the url but the url parser is going to take it as string anyway, but i can change it to 0 and 1

cornflow/tests/unit/test_instances.py
ggsdc: I had it in mind but I am trying to have the basic test first (coverage) and then test more complex things.

Adding to that list: test with different users

cornflow/tests/custom_test_case.py
ggsdc: Because in the row that we read from the database we have more fields that we do not have on the payload that is read from the file. For example:

- created_at.
- updated_at.
- deleted_at.

Or the id assigned to the objects that is being created

cornflow/tests/custom_test_case.py
ggsdc: See comment above

cornflow/tests/custom_test_case.py
ggsdc: Done

cornflow/tests/custom_test_case.py
ggsdc: Done

cornflow/tests/unit/test_instances.py
pchtsp: yes, indeed. do add it somewhere so we do not forget about it and anyone reading through the code can can implement it. thanks.

cornflow/endpoints/execution.py
pchtsp: ok i thought that in fact it had to be added to the endpoint in app.py in line https://github.com/baobabsoluciones/corn/blob/e5a780c5b6024f3cdae2d0c592376c99986eafe3/cornflow/app.py#L34 and there we could force it to be a binary. Right? Although I guess we would need to split the "post" from the "get" in different endpoints in order to only ask for an optional argument only in the post. We can leave it like this for now knowing it should be "documented" somewhere.

PR #17, opened 2021-01-15
Feature/json schema

hugolarzabal:
add a class managing json schemas


cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        """
+        loaded = len(self.jsonschema) > 0
+        if not loaded:
+            raise Warning("No jsonschema has been loaded in the SchemaManager class")
+        return loaded
+    
+    def validate_data(self, data_path):
+        """
+        Validate a json file according to the loaded jsonschema.
+
```
pchtsp, line 49: same as the other comment. Better to have two functions (one with a dictionary, one with a path)

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        :return: True if the data is valid and False if it is not.
+        """
+        self.is_schema_loaded()
+        
+        data = self.load_json(data_path)
+        try:
+            validate(data, self.get_jsonschema())
+            return True
+        except Exception as e:
```
pchtsp, line 60: does the jsonschema has specific exceptions you can put here to differentiate what is going wrong?

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+            print("json file not valid: ", e)
+            return False
```
pchtsp, line 62: maybe the return value could be a dictionary with errors. If it's empty: we're good. If it's not: we can see what's wrong. Do we get a detailed error report?

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        """
+        Transform a jsonschema into a dictionary and store it in self.schema_dict
+        
+        :return: The schema dictionary
+        """
+        jsonschema = self.get_jsonschema()
+        
+        if not self.is_schema_loaded():
+            return {"DataSchema": {}}
+        self.schema_dict = {"DataSchema": []}
```
pchtsp, line 74: why do we need to store it in `self.schema_dict`? is this operation "long" or do we expect it to be used many times with the same file? if we don't think it will be, we could just return this and never even store it (we could even delete the property)

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        """
+        Create flask marshmallow schemas from the jsonschema
+
+        :return: a dict containing the flask marshmallow schemas
+        """
+        self.jsonschema_to_dict()
+        self.dict_to_flask()
+        return self.flask_schemas
+        
+    def load_schema(self, path):
```
pchtsp, line 114: if you want to keep the self.schema_dict and the self.flask_schemas, you need to empty those each time you load a new schema.

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        for key, params in tuplist:
+            schema = ParameterSchema()
+            # this line validates the list of parameters:
+            params1 = schema.load(params, many=True)
+            result_dict[key] = gen_schema(key, params1, result_dict)
+    
+        self.flask_schemas = result_dict['DataSchema']
+        return self.flask_schemas
+    
+    def jsonschema_to_flask(self):
```
pchtsp, line 104: more than calling it "flask", I'd say marshmallow, right?

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        Load a json schema into self.jsonschema.
+
+        :param path: the path to the json file containing the schema.json
+
+        return the jsonschema
+        """
+        self.jsonschema = self.load_json(path)
+        return self.jsonschema
+
+    def print_schema_dict(self, path):
```
pchtsp, line 125: is it print or export?

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        """
+    
+        name = item[0]
+        content = item[1]
+    
+        schema = [dict(name=name,
+                      type=self._get_schema_name(name),
+                      many=("type" in content and content["type"] == "array"),
+                      required=True)]
+        self.schema_dict["DataSchema"] +=schema
```
pchtsp, line 175: are we assuming the name "DataSchema" inside this object? does it make sense? could we just assign this outside of this object? because this name is just a name we put to the pulp instance in Cornflow.

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        
+        :return A dict element for a schema_dict.
+        """
+        if required_list is None:
+            required_list = []
+        
+        name = item[0]
+        content = item[1]
+        if "type" not in content:
+            if "$ref" in content:
```
pchtsp, line 194: what happens when this inside if is false? do we get an error?

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        Transform an object item from the jsonschema in a dict for the schema_dict and update self.schema_dict.
+        In jsonschema objects are similar to python dict.
+        The object in jsonschema is in the following format:
+        "object_name": {"type":"object", "properties":{"field1": {...}, "filed2": {...}}, "required": ["field1]}
+        
+        The schema_dict object use the format:
+        {"schema_name": [{"name":"field1", "type": "field1_type", "many": False, "required":(True or False)}, ...]
+        
+        :param item: The jsonschema item (key, value)
+        The forma tof the item is: ("object_name", {"type":"object", "properties":{"a": {...}, "b": {...}})
```
pchtsp, line 229: The format of

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+    
+    def _get_type(self, json_type):
+        """
+        Translate the type between jsonschema and schema_dict.
+        
+        :param json_type: the type in jsonschema
+        
+        :return: the type in schema_dict.
+        """
+        if type(json_type) is list:
```
pchtsp, line 314: I think it's better to use `isinstance()`. It gets also the subclasses of list.

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+            if "null" in json_type:
```
pchtsp, line 315: I think it's more efficient to filter it directly instead of looking for it and then filtering.

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+    def _get_required(content):
+        """
+        Get the list of required name of it exist.
+        
+        :content: the dict which should have a "required" key.value
+        
+        :return: The required list or empty list.
+        
+        """
+        if "required" in content:
```
pchtsp, line 344: you can also do something like: `return content.get('required', [])`

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+        result_dict = {}
+        ordered = sort_dict(dict_params)
+        tuplist = sorted(dict_params.items(), key=lambda v: ordered[v[0]])
+        for key, params in tuplist:
+            schema = ParameterSchema()
+            # this line validates the list of parameters:
+            params1 = schema.load(params, many=True)
+            result_dict[key] = gen_schema(key, params1, result_dict)
+    
+        self.flask_schemas = result_dict['DataSchema']
```
pchtsp, line 101: also valid for `self.flask_schemas`, do we need to keep it or can we just recalculate at each time?

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+    def jsonschema_to_dict(self):
+        """
+        Transform a jsonschema into a dictionary and store it in self.schema_dict
+        
+        :return: The schema dictionary
+        """
+        jsonschema = self.get_jsonschema()
+        
+        if not self.is_schema_loaded():
+            return {"DataSchema": {}}
```
pchtsp, line 73: When is empty, couldn't you just return `self.schema_dict` ? Why is DataSchema is a dictionary here and later a list?

pchtsp (changes requested):
After the comments / corrections / etc. I will still wait until we get the tests for this code before approving.

cornflow/schemas/schema_manager.py
hugolarzabal: I think in general it is a good idea to do that. But here, I do not see how it could be a subclass (it is read from json) and why we should accept a subclass if it is.

cornflow/schemas/schema_manager.py
hugolarzabal: marshmallow is long and annoying to type. Technically a marshmallow schema is also a flask schema

cornflow/schemas/schema_manager.py
hugolarzabal: The last function creating the flask schema returns result_dict['DataSchema']

So the name DataSchema is not part of the final marshmallow schema. It is internal too the class and I think as good/bad as anything else.

cornflow/schemas/schema_manager.py
hugolarzabal: good question, I do not know what we should do if no type is given:
- raise an error here?
- make it a string by default?
- raise an error when transforming it into a flask schema?

cornflow/schemas/schema_manager.py
hugolarzabal: I working on it

cornflow/schemas/schema_manager.py
hugolarzabal: I use self.shema_dict in many functions without it being the return value of the function.

getting rid of this propery would change all the logic of the class.

I deleted the other properties.

cornflow/schemas/schema_manager.py
hugolarzabal: done

cornflow/schemas/schema_manager.py
```diff
@@ -0,0 +1,368 @@
...
+from jsonschema import validate
+from copy import deepcopy
+from genson import SchemaBuilder
+from ..schemas.schema_dict_functions import gen_schema, ParameterSchema, sort_dict
+from .constants import JSON_TYPES
+
+
+class SchemaManager:
+    
+    def __init__(self, schema_path=None):
```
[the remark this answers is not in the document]
hugolarzabal: done

cornflow/schemas/schema_manager.py
hugolarzabal: done

cornflow/schemas/schema_manager.py
hugolarzabal: I changed it. I am not sure that returning [] when the file is valid is right.

cornflow/schemas/schema_manager.py
hugolarzabal: done, it returns a list of errors

cornflow/schemas/schema_manager.py
hugolarzabal: done

cornflow/schemas/schema_manager.py
hugolarzabal: done

cornflow/schemas/schema_manager.py
hugolarzabal: done

cornflow/schemas/schema_manager.py
hugolarzabal: done

cornflow/schemas/schema_manager.py
hugolarzabal: export, done

cornflow/schemas/schema_manager.py
hugolarzabal: done

cornflow/schemas/schema_manager.py
hugolarzabal: Now there are 3 functions get_validation_erros, validate_data and validate_file

cornflow/schemas/schema_manager.py
pchtsp: ok for now we leave as it is. But I would like to replace the "property" with a call to the function that produces it.

cornflow/schemas/schema_manager.py
pchtsp: haha you could put, I don't know, `marsh` ? Ok, we leave it with flask.

cornflow/schemas/schema_manager.py
pchtsp: I have no idea what the code does actually, I just noticed the ifs. If it's wrong (unexpected): we should raise an error saying "incorrect jsonschema" and that the object `name` is missing a type. I've just checked that "raise TypeError()" exists so you could use that haha.

PR #18, opened 2021-01-18
Feature/airflow status and mps file

pchtsp:
This PR has some new changes.

**Beware**. One of this changes may break something. Now the execution status endpoint ("execution-status") now returns a json with the format `{'status': 'state'}`. I'm not sure if this was the case before.

Changes:

1. Cornflow now asks the status to airflow in case it has not finished.
2. New cornflow endpoint to send an mps. With some tests.
3. A couple tests on the cornflow-client.
4. Tests now use sqlite instead of postgres.

As a side comment I think we're going to have to change the name and type of the "finished" column in the execution table. That way we can store "finished" or "error". In both of these cases it should return the status to the user and not ask airflow for more info.


README.rst
```diff
@@ -332,31 +330,25 @@ Appended in this repository are three more docker-compose files for different ki
...
-    sudo su - postgres
-    psql -c "create database cornflow_test"
-    exit
-
 Then you have to run the following commands::
 
-    export FLASK_APP=cornflow.app
     export FLASK_ENV=testing
 
 Finally you can run all the tests with the following command::
```
ggsdc, line 338: Why do do you not want to run the test directly with coverage?

cornflow/endpoints/execution.py
```diff
@@ -60,26 +59,33 @@ def post(self):
...
-        # To send the absolute url:
-        # url_for(endpoint_name, _external=True)
-        airflow_client = Airflow(current_app.config['AIRFLOW_URL'],
-                                 current_app.config['AIRFLOW_USER'],
-                                 current_app.config['AIRFLOW_PWD'])
-        if not airflow_client.is_alive():
+
+        # We now try to launch the task in airflow
+        af_client = Airflow(**airflow_conf)
+        if not af_client.is_alive():
```
ggsdc, line 76: Maybe this error message should be different?

ggsdc: The client should not need to know that Airflow is behind everything

cornflow/endpoints/execution.py
```diff
@@ -60,26 +59,33 @@ def post(self):
...
             return {'error': "Airflow is not accessible"}, 400
         execution_id = result[0][self.primary_key]
         try:
-            airflow_client.run_dag(execution_id, current_app.config['CORNFLOW_URL'])
+            response = af_client.run_dag(execution_id, config['CORNFLOW_URL'], dag_name='solve_model_dag')
         except AirflowApiError as err:
```
ggsdc, line 81: Same here as the above message

cornflow/endpoints/execution.py
```diff
@@ -159,11 +165,24 @@ def get(self, idx):
...
-        if not status:
-            # Here we should call airflow to check solving status
-            pass
+        airflow_conf = dict(url=current_app.config['AIRFLOW_URL'],
+                            user=current_app.config['AIRFLOW_USER'],
+                            pwd=current_app.config['AIRFLOW_PWD'])
 
-        return {'finished': status}, 200
+        self.user_id, self.admin, self.super_admin = Auth.return_user_info(request)
+        execution = ExecutionModel.get_one_execution_from_user(self.user_id, idx)
```
ggsdc, line 173: Maybe the status variable is not needed. execution.finished is by defect false, so you could do if execution.finished instead

cornflow/endpoints/instance.py
```diff
@@ -117,17 +122,55 @@ def delete(self, idx):
...
+    def post(self):
         """
-        Not implemented!
 
         :param file:
         :return:
         :rtype: Tuple(dict, integer)
         """
-        return {}, 501
+        self.user_id, self.admin, self.super_admin = Auth.return_user_info(request)
```
ggsdc, line 138: How is this form actually sent?

ggsdc:
I think everything is alrright, just some minor comments. Check them and we can merge.

ggsdc:
I've just see that my comments in this page are not correctly seen as the show the above code, not the below code, that is the one that I am always referencing, please check the comments in the files changed tab.

cornflow/endpoints/execution.py
```diff
@@ -159,11 +165,24 @@ def get(self, idx):
...
+        status = execution.finished
+        if status:
+            return {'status': 'finished'}, 200
+        dag_run_id = execution.dag_run_id
```
ggsdc, line 177: This is what I was refering on the other comment, but come to think of it if we want to change this column, we should keep the following in mind:

- We should have a column with a meaningful way to show the status of the run: running, failed, finished correctly. These are the three states that I can think as of right now.
- We should have an extra column, that can have null values, with a meaningful description of what happened (mainly to store failed messages).

Does this make sense? or should be simplified? Because if for example the error is that Airflow is not available, that is not going to be stored on the log_json or log_text fields.

README.rst
pchtsp: good question. I've always seen the tests being run by `unittest` directly in other places. Coverage is to get the coverage, right? It seems it uses `unittest` under the hood so in the end is the same thing. When we need the coverage, then it makes sense to run `coverage` (as in the last command).

cornflow/endpoints/execution.py
pchtsp: Good point. For debugging is useful but maybe the client should not know this. One option is saying something like "scheduler of jobs is not working, check Airflow is working" or just "job schedule is unavailable." Any other idea is welcome.

cornflow/endpoints/execution.py
pchtsp: yes, this is maybe too technical. Useful for debugging but we should change it for something less cryptic. Or maybe put both (the airflow error but first introduce something about the "scheduler")

pchtsp: yes, this is maybe too technical. Useful for debugging but we should change it for something less cryptic. Or maybe put both (the airflow error but first introduce something about the "schedul

cornflow/endpoints/execution.py
pchtsp: Yes, I will change this. I think we returned the `status` boolean variable before and I did not change it.

cornflow/endpoints/execution.py
pchtsp: I agree with the two column approach. The "failed", "running" and "finished" could be three integer codes: -1, 0 and 1 respectively. And then we have the additional column with the error message. It reminds me of aimms pro.

pchtsp: I agree with the two column approach. The "failed", "running" and "finished" could be three integer codes: -1, 0 and 1 respectively. And then we have the additional column with the error message. It reminds me of aimms pro.

cornflow/endpoints/instance.py
pchtsp: You can check the [cornflow-client tests](https://github.com/baobabsoluciones/corn/blob/feature/airflowStatus_and_mpsFile/cornflow/tests/unit/test_cornflowclient.py) or the [flask tests](https://github.com/baobabsoluciones/corn/blob/feature/airflowStatus_and_mpsFile/cornflow/tests/unit/test_instances_file.py).

cornflow/endpoints/execution.py
ggsdc: Ok, i will do this change in a new branch. Then we have to modify the way the status is checked and do:

```
status = execution.state == 1
if status:
    ...
```

or: 
```
if execution.state == 1:
    ...
```

Because i guess you want -1 for failure, 0 for running and 1 for finished

PR #20, opened 2021-01-21
Modified state logging of the execution.

ggsdc:
Added migration


cornflow/endpoints/dag.py
```diff
@@ -38,6 +38,9 @@ def post(self, idx):
         # because we do not want to store airflow's user:
         req_data['user_id'] = execution.user_id
         req_data['finished'] = True
+        # TODO: here I guess that the run is performed without problems so state should be 1
+        req_data['state'] = 1
```
pchtsp, line 42: We should define the constants in some other file, and use them here, such as EXEC_STATE_CORRECT=1, EXEC_STATE_ERROR=-1, EXEC_STATE_RUNNING=0, EXEC_STATE_STOPPED=-2, EXEC_STATE_UNKNOWN=-3 etc.

cornflow/endpoints/execution.py
```diff
@@ -171,17 +172,28 @@ def get(self, idx):
 
         self.user_id, self.admin, self.super_admin = Auth.return_user_info(request)
         execution = ExecutionModel.get_one_execution_from_user(self.user_id, idx)
-        if execution.finished:
+        if execution.state == 1:
```
pchtsp, line 175: i think it's better to just return the state as "status" instead of a text. We could also return the `state_message` but the `state` makes more sense.
I mean:
```return dict(status=execution.state), 200```
Alternatively (which I see you have done for the others): we could do:
```return dict(status=execution.state_message), 200```
At least that way we do not duplicate the string with the error.

File: cornflow/endpoints/execution.py [hunk 3]
```diff
@@ -171,17 +172,28 @@ def get(self, idx):
...
             return {'status': 'finished'}, 200
         dag_run_id = execution.dag_run_id
         if not dag_run_id:
+            execution.state = -1
```
pchtsp, line 179: here, I would use EXEC_STATE_UNKNOWN instead of "ERROR" because it did not even start the run. We can maybe create an error specific to that too (EXEC_STATE_ERROR_START) or something.

cornflow/endpoints/execution.py
```diff
@@ -171,17 +172,28 @@ def get(self, idx):
...
+            execution.state_message = 'The execution has no dag_run associated'
+            execution.save()
             return {'error': 'The execution has no dag_run associated'}, 400
         af_client = Airflow(**airflow_conf)
         if not af_client.is_alive():
+            execution.state = -1
```
pchtsp, line 185: maybe this one is the same as ERROR_START. Ideally, if the state_message changes, the error should be different.

cornflow/endpoints/execution.py
```diff
@@ -171,17 +172,28 @@ def get(self, idx):
...
+            execution.state_message = 'Airflow is not accessible'
+            execution.save()
             return {'error': "Airflow is not accessible"}, 400
         try:
             response = af_client.get_dag_run_status(dag_name='solve_model_dag', dag_run_id=dag_run_id)
         except AirflowApiError as err:
+            execution.state = -1
+            execution.state_message = "Airflow responded with an error: {}".format(err)
             return {'error': "Airflow responded with an error: {}".format(err)}, 400
```
pchtsp, line 194: here you need to save the `execution`.

cornflow/models/execution.py
```diff
@@ -63,7 +64,11 @@ class ExecutionModel(BaseAttributes):
     execution_results = db.Column(JSON, nullable=True)
     log_text = db.Column(TEXT, nullable=True)
     log_json = db.Column(JSON, nullable=True)
-    finished = db.Column(db.Boolean, server_default=expression.false(), default=False, nullable=False)
+    state = db.Column(db.SmallInteger, default=0, nullable=False) # IntegerRangeField
```
pchtsp, line 67: it makes sense to have the default at 0? I would import the constants here too and assign the constant with a name. I think it's better to assign something like "UNKNOWN_ERROR" by default.

cornflow/schemas/execution.py
```diff
@@ -35,6 +35,8 @@ class ExecutionSchema(Schema):
     log_text = fields.Str(dump_only=True)
     log_json = fields.Nested(LogSchema, dump_only=True)
     finished = fields.Boolean(required=False)
+    state = fields.Int(validate=validate.Range(min=-1, max=1), required=False)
```
pchtsp, line 38: this too, will need to be edited if we add errors. Maybe we should leave it open so it doesn't fail when we add a new error.

migrations/versions/e1a50dae1ac9_.py
```diff
@@ -18,50 +19,55 @@
...
-    sa.Column('id', sa.Integer(), nullable=False),
-    sa.Column('name', sa.String(length=128), nullable=False),
-    sa.Column('email', sa.String(length=128), nullable=False),
-    sa.Column('password', sa.String(length=128), nullable=True),
-    sa.Column('admin', sa.Boolean(), server_default=sa.text('false'), nullable=False),
-    sa.Column('super_admin', sa.Boolean(), server_default=sa.text('false'), nullable=False),
-    sa.PrimaryKeyConstraint('id'),
-    sa.UniqueConstraint('email')
-    )
+                    sa.Column('created_at', sa.DateTime(), nullable=False),
```
pchtsp, line 24: I know this is automatic but it's weird that they have that much indentation...

pchtsp (changes requested):
I think we could already add some tests, even if we do not have airflow yet in the tests. Specially the cases when airflow throws an error because it doesn't exist and when running the execution without airflow.

cornflow/endpoints/dag.py
ggsdc: Ok, I can do that, but for now, the state can only take the following values. -1, 0 and 1. I have set up a constraint in the schema valdiation and in the database, if want to have more codes I have to change also those constraints

cornflow/schemas/execution.py
ggsdc: I prefer to modify it every time we add a new error, as there is also a database constraint on the value so that we know which values we can have and which we cant. I can define it in the constant file, so that when we add a new error and its code it is automatically changed everywhere

migrations/versions/e1a50dae1ac9_.py
ggsdc: It the indentation changed by pycharm in the file. It is its standard formatting

cornflow/models/execution.py
ggsdc: The moment the execution is created, by the default is going to be run, so i guess the default state should be running.

cornflow/schemas/execution.py
pchtsp: yes, better to have it as a defined constant in the `constants.py`. But then you have to change it in the model too and tie it to the `constant.py` file? seems complicated...

cornflow/models/execution.py
pchtsp: yeah but if an error happens somewhere that avoids detection and so an update of the error, then it will show as running for ever. Well, we can leave it like that, yes: it also makes sense.

cornflow/schemas/execution.py
```diff
@@ -35,6 +36,9 @@ class ExecutionSchema(Schema):
     log_text = fields.Str(dump_only=True)
     log_json = fields.Nested(LogSchema, dump_only=True)
     finished = fields.Boolean(required=False)
+    state = fields.Int(validate=validate.Range(min=MIN_EXECUTION_STATUS_CODE, max=MAX_EXECUTION_STATUS_CODE),
+                       required=False)
+    state_message = fields.Str(required=False)
     created_at = fields.DateTime(dump_only=True)
```
ggsdc, line 42: So @pchtsp here you commented that It could be complicated.

I'll explain how it is now and I can change it if you want.

Codes are defined in shared/const.py file. From that file are read automatically in the models/execution.py and here as well. If you where to add a new code and modify the min and max code, here the validation in the schema would work.

But, this is setup as well as a database constraint in the corresponding migration file. If you where going to add new codes you would have to create a manual migration (create the file by hand) deleting the constraint and creating it again with the new values. This part is the one that seems more complicated (IMHO) and maybe is the one that we should delete.

Tell me what you think.

.gitignore
```diff
@@ -108,6 +108,7 @@ env/
 venv/
 cfvenv/
 afvenv*/
+lenv*/
 ENV/
```
ggsdc, line 112: Here i just added another virtualenvironmewnt for wsl in my local machine and added this so it does not ask me all the time to add these files

ggsdc:
I ran the test in my local machine and we are up to a 87% of coverage. I am going to add some unittest to cover some of the new funtionalities of this PR

PR #53, opened 2021-05-26
More scenarios endpoints

ggsdc:


cornflow/schemas/case.py
```diff
@@ -51,3 +53,17 @@ class CaseFromInstanceExecution(Schema):
...
+
+    id = fields.Str(required=True)
+    schema = fields.Str(required=True)
+
+
+class CaseEditRequest(Schema):
+    name = fields.String()
+    description = fields.String()
+    schema = fields.String()
+    path = fields.Str()
```
pchtsp, line 69: does it make sense to edit the `path` property here? in theory that should be done through the `move` endpoint, isn't it?

cornflow/schemas/case.py
```diff
@@ -51,3 +53,17 @@ class CaseFromInstanceExecution(Schema):
     name = fields.Str(required=True)
     description = fields.Str()
     path = fields.Str(required=True)
+
+
+class CaseToLive(Schema):
```
pchtsp, line 58: I'd change this name to `CaseToInstanceResponse` or something like that.

cornflow/endpoints/case.py
```diff
@@ -217,9 +247,63 @@ def get(self, idx):
...
+        if schema is None:
+            return self.post_list(payload)
+
+        if schema == "pulp" or schema == "solve_model_dag":
+            validate_and_continue(DataSchema(), payload["data"])
+            return self.post_list(payload)
+
+        config = current_app.config
+        marshmallow_obj = get_schema(config, schema)
+        validate_and_continue(marshmallow_obj(), payload["data"])
```
pchtsp, line 307: we have this code in several places in the codebase (several endpoints). Maybe we should think on creating a single function that does this and call it from each. I'm not sure / convinced.

cornflow/schemas/case.py
```diff
@@ -1,5 +1,7 @@
 from marshmallow import fields, Schema
 
+from .common import QueryFilters as QueryFiltersCase
+
```
pchtsp, line 5: so it doesn't appear it's not being used I edited this line in the instance and execution schema so it looks like this:

```
class QueryFiltersCase(QueryFilters)
    pass
```

cornflow/tests/unit/test_cases.py
```diff
@@ -272,3 +272,76 @@ def test_get_nonexistent_instance(self):
         self.get_one_row(
             self.url + "500" + "/", {}, expected_status=404, check_payload=False
         )
+
+    def test_update_one_instance(self):
```
pchtsp, line 276: Copying and pasting code looks bad: specially if the names of the methods are not even modified (instance?)

If we're re-using code from the instances, I'd create a meta class for testing and use it for both. For ways of reusing tests for two classes, check `BaseTestCases.ListFilters` in the `custom_test_case.py` file. We use those tests in cases, instances and executions.

cornflow/endpoints/case.py
ggsdc: I can think about it. Maybe it makes sense to generalize it if we also start to validate the config data for the executions.

ggsdc (commented):
The comments have been covered

ggsdc:
@pchtsp The changes that you have petitioned should be covered now.
I have started some generalization of some test cases, but have decided to postpone it for later as I want to finish other tasks first, but I have more ideas on how to finish the generalization.

PR #54, opened 2021-05-27
Feature/scenarios jsonpatch

ggsdc:
Added endpoint to modify case data with json patch


cornflow/models/meta_model.py
```diff
@@ -76,6 +81,23 @@ def update(self, data):
...
+    def patch(self, data):
+        """
+
+        :param dict data:
+        :type data:
+        """
+        try:
+            self.data = jsonpatch.apply_patch(self.data, data.get("patch"))
+        except jsonpatch.JsonPatchConflict as e:
+            raise InvalidPatch(payload={"message": str(e)})
```
pchtsp, line 93: I think we usually pass the error message in the error attribute of the error and not in the payload. We don't use the payload except for AirflowErrors (not sure why). Check `InvalidUsage(error=val_err.normalized_messages())` in `endpoints/instance.py`. You can do: `InvalidPatch(error=str(e))` or `InvalidPatch(error=e.normalized_messages())` if that is available.

cornflow/tests/unit/test_cases.py
```diff
@@ -307,3 +314,66 @@ def test_case_does_not_exist(self):
...
+class TestCaseJsonPatch(CustomTestCase):
+    def setUp(self):
+        super().setUp()
+        self.payload = self.load_file(CASE_PATH)
+        self.model = CaseModel
+        self.case_id = self.create_new_row(CASE_URL, self.model, self.payload)
+        self.payloads = [self.load_file(f) for f in CASES_LIST]
+        self.items_to_check = ["name", "description", "path", "schema"]
+        self.url = CASE_URL
+        self.patch = {
```
pchtsp, line 327: we're validating jsonpatch with jsonpatch? I think we should put the difference inside explicitly in the test so we know what we're expecting as a json. It should not be too big (we can generate it once, validate it and paste it as json/dict)

pchtsp (commented):
Thanks! . There's just two minor changes.

cornflow/models/meta_model.py
ggsdc: Ok, the error message is quite huge but i did not try with the normalized_message. Will try it out and pass it as the error, not the message.

cornflow/tests/unit/test_cases.py
ggsdc: You are right, i should have a json-patch file on the data folder and test that json-patch creates that file and if we send that file it works.

I think we can keep these test nonetheless and create new ones reading the json-patch from a file

cornflow/models/meta_model.py
ggsdc: Ok, so the exception raised by jsonpatch does not have an implementation of normalized message, so I am going to change it so it does not give back the full error in the messages attribute and we can think later if we want to capture the error and how.

PR #55, opened 2021-05-28
Case compare endpoint

ggsdc:
Added endpoint to compare two cases given its id.
Added test to test its behaviour.
Added data files for the unit tests


cornflow/endpoints/case.py
```diff
@@ -315,3 +318,56 @@ def post(self, idx):
...
+        :return:an object with the instance or instance and execution ID that have been created and the status code
+        :rtype: Tuple (dict, integer)
+        """
+        if idx1 == idx2:
+            raise InvalidData("The case identifiers should be different", 400)
+        case_1 = self.model.get_one_object_from_user(self.get_user(), idx1)
+        case_2 = self.model.get_one_object_from_user(self.get_user(), idx2)
+
+        if case_1 is None:
+            raise ObjectDoesNotExist("The first case does not exist")
```
pchtsp, line 358: I prefer something like "You do not have access to the first case". I'm not sure if we should give back the code of the case...

cornflow/endpoints/case.py
```diff
@@ -315,3 +318,56 @@ def post(self, idx):
...
+        elif case_2 is None:
+            raise ObjectDoesNotExist("The first case does not exist")
+
+        data = kwargs.get("data", 1)
+        solution = kwargs.get("solution", 1)
```
pchtsp, line 363: it's a dumb change but I prefer `data = kwargs.get("data", True)` it explains better that it's boolean.

cornflow/tests/data/full_case_raw.json
```diff
@@ -0,0 +1,197 @@
+{
```
pchtsp, line 1: I prefer to import the files directly from cornflow_client. cornfllow_client is a requirement in any case. This helps later in the rare cases where the schemas change: we only change them in cornflow_client. It has happened before and it was horrible searching for all libraries where this file was present.

pchtsp (commented):
very small changes

cornflow/endpoints/case.py
ggsdc: It does not have to mean that you do not have access, it can also mean that the case does not exist and the id is wrong. But I see your point that it could be both so I should change the message

cornflow/endpoints/case.py
ggsdc: Ok, as in the url is 0 or 1 I decided to put as default (remember that if is not specified should be True. 

But I can change it.

cornflow/tests/data/full_case_raw.json
ggsdc: Ok, then I will change it to import both files from cornflow client.

requirements.txt
```diff
@@ -36,7 +36,7 @@ Werkzeug==1.0.1
 flask-cors
 pulp
 flask-apispec
```
ggsdc, line 38: This change is so that it takes the change to the function in the PR in cornflow-cllient (https://github.com/baobabsoluciones/cornflow-client/pull/10), once that PR is accepted I will update this PR to point back to the cornflow-client master

ggsdc:
@pchtsp This should be ready to merge. It also fixes an error that @dlopbar found in the deletion of a case, so doing the merge should fix the problem he is encountering.

PR #56, opened 2021-06-01
take out user requirement to ask for schemas.

pchtsp:
This way we can ask for the schema when the web app opens (even if no user is logged in).


ggsdc (commented):
Seem ok to me

PR #57, opened 2021-06-01
Now patch changes solution too

ggsdc:


cornflow/models/meta_model.py
```diff
@@ -87,14 +87,25 @@ def patch(self, data):
...
+                except jsonpatch.JsonPatchConflict:
+                    raise InvalidPatch()
+                except jsonpatch.JsonPointerException:
+                    raise InvalidPatch()
+            else:
+                setattr(self, key, item)
+
+        for key, item in vars(self).items():
+            if "hash" in key:
+                setattr(self, key, hash_json_256(getattr(self, key.split("_")[0])))
```
pchtsp, line 107: Not sure what you want to do but this code is almost uncomprehensible. I do not like particularly:

1. the use of setattr, getattr vars(self). We need to stop using that. ([I'm not the only one who says that](https://google.github.io/styleguide/pyguide.html#219-power-features))
2. the amount of indentation is too large

Is this an attempt to do `data_patch` and `solution_patch`? Can't we just define a helper function and call it two times? once for "data_patch", another for "solution_patch" with some arguments? Maybe there's something I do not understand.

Also: are we using "patch" or "data_patch"? Because I see we changed it in the tests but here it's still "patch".

pchtsp (changes requested):
On top of the comment for the `meta_model.py`:

1. Why do we have that function in the meta_model.py? We can only patch cases so I do not see why we shouldn't have that logic in the model/case.py. Also for patch. If later we want to have the exact same logic for instances and executions: we can move it.
2. Unless I'm mistaken, the tests do not include `data_solution`. We need to check that works too.

cornflow/models/meta_model.py
ggsdc: Ok, I will think a way to make the code simplier and readable withouth using getattr setattr or vars.

It has to be name of the field_patch.

ggsdc:
>     2\. a_s

Thought it was good to have it on meta_model so we can reuse it later one, but yes it could be on models/case.py instead. I think some of the test cover the solution has well, let me check.

ggsdc (commented):
Changes done

pchtsp (commented):
we're still missing the test for the solution_patch if I'm not mistaken (i do not see any input test data that includes the word "solution_patch")

ggsdc:
> 
> 
> we're still missing the test for the solution_patch if I'm not mistaken (i do not see any input test data that includes the word "solution_patch")

As I said, the test test_get_patch_and_apply compares two cases with data and solution and then applies the patch generated from both cases to the first one. Is in the TestCaseCompare suite, not in the jsonpatch one, but should be enough for the merge, we can add more test later. I can open an issue if you want.

PR #62, opened 2021-06-10
Feature/ldap

ampuerin:
Documentation of new user access control feature for cornflow
Remove from docker-compose files using the official image from docker-hub (now build image on run)
New entrypoint scripts for docker images (ldap integration)
Openldap populate db file in airflow_config folder (ldapbootstrap.ldif) 
Use airflow 2.0.2 in all deployments


airflow_config/Dockerfile
```diff
@@ -26,22 +26,28 @@ RUN apt update -y && apt-get install -y --no-install-recommends \
         ldap-utils \
         libffi6 \
         libffi-dev \
+        libldap2-dev \
```
pchtsp, line 29: Are we installing ldap libraries "just in case" we want to use ldap with airflow too?

airflow_config/scripts/initairflow.sh
```diff
@@ -18,6 +18,8 @@
 : "${AIRFLOW_ROLE:="Admin"}"
```
pchtsp, line 18: I wonder if maybe we should migrate these deployment scripts to python. It's possible we would be able to reduce the size and make them more flexible. What do you think?

initapp.sh
```diff
@@ -10,21 +10,36 @@
 : "${FLASK_ENV:="development"}"
```
pchtsp, line 10: this is the other file I'm wondering if we would like to migrate to python. What do you think?

pchtsp (commented):
I have only three questions.

PR #64, opened 2021-06-10
User role model ppost endpoint and tests

ggsdc:
User role endpoint checks if role already assigned or if disabled and acts accordingly.
Fixed tests.
Added more docstrings


cornflow/commands/commands.py
```diff
@@ -55,36 +69,59 @@ def get_options(self):
...
 
         :param str email: the email for the admin user
         :param str password: the password for the admin user
+        :param int verbose: verboseof the command
         :return: a boolean if the execution went right
         :rtype: bool
         """
-        return create_user_with_role(email, password, "serviceuser", SERVICE_ROLE)
+        if email is None or password is None:
+            print("Missing required arguments")
```
pchtsp, line 91: do we want to continue if both arguments are not set? I'm not sure to understand why put default arguments if we're going to check if they are None.

cornflow/commands/commands.py
```diff
@@ -55,36 +69,59 @@ def get_options(self):
...
+            ),
         )
 
-    def run(self, email, password):
+    def run(self, email=None, password=None, verbose=0):
         """
         Method to run the command and create the service user
 
         :param str email: the email for the admin user
         :param str password: the password for the admin user
```
pchtsp, line 85: it says "admin". should say "service"

cornflow/commands/commands.py
```diff
@@ -55,36 +69,59 @@ def get_options(self):
...
+        :param int verbose: verbose of the command
         :return: a boolean if the execution went right
         :rtype: bool
         """
-        return create_user_with_role(email, password, "admin", ADMIN_ROLE)
+        if email is None or password is None:
+            print("Missing required arguments")
+        verbose = int(verbose)
```
pchtsp, line 123: you can pass a type argument to the Option I think so you don't have to cast to int: https://flask-script.readthedocs.io/en/latest/#flask_script.Option

cornflow/models/roles.py
```diff
@@ -173,6 +186,42 @@ def is_service_user(user_id):
...
+        :param user_id: id of the specific user
+        :param role_id: id of the specific role
+        :return: a boolean if the user has the role assigned but disabled
+        :rtype: bool
+        """
+        user_role = UserRoleModel.query.filter(
+            UserRoleModel.user_id == user_id,
+            UserRoleModel.role_id == role_id,
+            UserRoleModel.deleted_at != None,
+        ).first()
```
pchtsp, line 219: I think you can chain queries in sqlalchemy like:
```
user_role = UserRoleModel.get_one_user_role(user_id, role_id).query(UserRoleModel.deleted_at != None)
```
or something similar. Check the examples we have in `meta_model.py` (`get_all_objects`)

cornflow/models/roles.py
```diff
@@ -173,6 +186,42 @@ def is_service_user(user_id):
...
+        if user_role is None:
+            return False
+        else:
+            return True
+
```
pchtsp, line 224: better yet: `return user_role is not None`

cornflow/tests/unit/test_roles.py
```diff
@@ -311,22 +335,32 @@ def test_delete_user_role_authorized_user(self):
...
-        user_role = self.create_role(user_response.json["id"], role)
+        planner_role = UserRoleModel.query.filter_by(
+            user_id=user_response.json["id"]
+        ).first()
+
+        self.create_role(user_response.json["id"], role)
+
         self.client.delete(
-            self.url + str(user_role.id) + "/",
+            self.url + str(planner_role.id) + "/",
```
pchtsp, line 352: shouldn't it be `planner_role.role_id`?

`planner_role.id` returns the id of the row, not the id of the role. If I understand correctly.

cornflow/models/roles.py
ggsdc: I will try, but as` UserRoleModel.get_one_user_role` has a` first()` at the end I do not know if we can chain it

cornflow/commands/commands.py
ggsdc: Ok, will modify it

cornflow/commands/commands.py
ggsdc: Will fix

cornflow/models/roles.py
ggsdc: As I tought, having the `first()` doesnt allow to chain queries.

cornflow/tests/unit/test_roles.py
pchtsp: mmm i don't think it makes sense to have the endpoint require the id of the assignment. It should require "role+user". The internal assignment id should be unknown to the client. I think then you should change the endpoint to require user_id/role_id for the put / getOne / delete.

cornflow/commands/commands.py
```diff
@@ -53,38 +67,80 @@ class CreateServiceUser(Command):
...
+                dest="password",
+                help="Service user password",
+                type=str,
+            ),
+            Option(
+                "-v",
+                "--verbose",
+                dest="verbose",
+                help="Verbose for the command. 0 no verbose, 1 full verbose",
+            ),
```
pchtsp, line 83: you're missing the type here

cornflow/commands/commands.py
pchtsp: But I still don't understand why the arguments have default values. Why do we allow default values if we're not going to let the function work without them. Shouldn't we just take out the default values (as it was before?)

cornflow/commands/commands.py
```diff
@@ -144,11 +227,32 @@ def run(self):
...
-    def run(self):
+    def get_options(self):
+        return (
+            Option(
+                "-v",
+                "--verbose",
+                dest="verbose",
+                help="Verbose for the command. 0 no verbose, 1 full verbose",
+                type=int,
+            ),
```
pchtsp, line 245: in fact inf you're using this in all functions... couldn't you just create it once at the top of the file such as:
```
verbose_option = Option("-v", ...)
```
and then just pass `verbose_option` everywhere?

cornflow/tests/unit/test_roles.py
ggsdc: Ok, will change it so it expects the user_id and the role_id

cornflow/commands/commands.py
ggsdc: Will fix it

cornflow/commands/commands.py
ggsdc: Will try

cornflow/commands/commands.py
ggsdc: The default values were because PyCharm was raising an error that could be fixed assigning default values. See the image:

image

Thats why I added the default values, but I can take them out if you want

PR #67, opened 2021-06-14
Change log in to username

ggsdc:


migrations/versions/4aac5e0c6e66_.py
```diff
@@ -0,0 +1,79 @@
+"""
```
ggsdc, line 1: This migration file does not need changes

migrations/versions/ca449af8034c_.py
```diff
@@ -0,0 +1,75 @@
+"""
```
ggsdc, line 1: This migration file could be changed to not uses the engine name

ggsdc:
The integration test do not pass as we need to change the cornflow-client library

PR #71, opened 2021-07-16
added more tests to dags and commands.

pchtsp:
we're having issues with dag executions and with some special users. I've added more tests in both places.


ggsdc:
Everything seems alright to me.

ggsdc (approved):

Issue #70, opened 2021-07-15
A service user cannot post a manual dag


pchtsp:
Neither can a normal user.

See cornflow/endpoints/dag.py:DAGEndpointManual

It has `ROLES_WITH_ACCESS = [ADMIN_ROLE, SERVICE_ROLE]` But it does accept the request from a `SERVICE_ROLE` user. It does not accept a planner user, I guess. It should allow, at least, the `SERVICE_ROLE` and probably, the `PLANNER_ROLE`.

ggsdc:
Ok so for the service user doesn't work because the service user does not have permission to post in any endpoint:

```
BASE_PERMISSION_ASSIGNATION = [
    (VIEWER_ROLE, GET_ACTION),
    (PLANNER_ROLE, GET_ACTION),
    (PLANNER_ROLE, PATCH_ACTION),
    (PLANNER_ROLE, POST_ACTION),
    (PLANNER_ROLE, PUT_ACTION),
    (PLANNER_ROLE, DELETE_ACTION),
    (ADMIN_ROLE, GET_ACTION),
    (ADMIN_ROLE, PATCH_ACTION),
    (ADMIN_ROLE, POST_ACTION),
    (ADMIN_ROLE, PUT_ACTION),
    (ADMIN_ROLE, DELETE_ACTION),
    (SERVICE_ROLE, GET_ACTION),
    (SERVICE_ROLE, PATCH_ACTION),
    (SERVICE_ROLE, PUT_ACTION),
    (SERVICE_ROLE, DELETE_ACTION),
]
```

I could add an exception in `EXTRA_PERMISSION_ASSIGNATION = [(VIEWER_ROLE, PUT_ACTION, "user-detail")]` and that should fix it. This same exception could be added to the `PLANNER_ROLE`.

PR #74, opened 2021-07-19
Service role and planner role can post manual dags now

ggsdc:
This PR solves issue #70


cornflow/tests/unit/test_dags.py
```diff
@@ -23,14 +26,30 @@ def test_manual_dag(self):
...
-        #     payload=payload_to_send,
-        #     check_payload=False,
-        #     token=token,
-        # )
-        # print(idx)
+        idx = self.create_new_row(
+            url=DAG_URL,
+            model=self.model,
+            payload=payload_to_send,
+            check_payload=False,
```
pchtsp, line 33: dumb question: could we be able to validate with `check_payload=True`? I had it at `False` in the commented code but I'm not sure if it was because we could not check or because I was lazy.

ggsdc: We could, but we have to check which keys we are going to compare, as the sent data and the response data do not have the same keys. Let me check if I can change it

ggsdc:
Now it checks the payload, I just had to set up which information has to be compared as the payload has the data but the response only has the data_hash.

Also I had to check the state to EXEC_STATE_MANUAL so that the test passes. Check if everything is alright

Issue #75, opened 2021-07-19
PostgreSQL does check if foreign key exists before inserting new data to database but sqlite does not.


ggsdc:
We should create a new github actions flow where we test the whole application with a postgresql database engine before making more changes.

Issue #77, opened 2021-07-19
Review why test doen't work with postgres backend


ggsdc:
Test `test_new_role_authorized_user` doesn't work with a postgresql engine. Fix the primary key sequence and autoincrement.

ggsdc:
This is caused because when we create the roles we insert the data with a given key and we do not allow PostgreSQL to set its own auto increment key, so the sequence does not get updated and when the test tries to create a new role has a primary key violation exception error.

This can be solved updating the sequence after the roles have been created by the command. This SQL Statement should be tried in SQLite to check if it works too.

More info: https://stackoverflow.com/questions/244243/how-to-reset-postgres-primary-key-sequence-when-it-falls-out-of-sync

PR #78, opened 2021-07-19
Added flow with postgresql.

ggsdc:
One test has been disabled for now

This partially covers issue #75

### DIFF
--- a/.github/workflows/run_tests_postgresql.yml
+++ b/.github/workflows/run_tests_postgresql.yml
@@ -41,6 +41,4 @@ jobs:
       run: python -m unittest discover -s cornflow.tests.unit
       env:
         FLASK_ENV: testing
-        POSTGRES_HOST: localhost
-        POSTGRES_PORT: 5432
         DATABASE_URL: postgres://postgres:postgresadmin@127.0.0.1:5432/cornflow-test
--- a/.github/workflows/run_tests_postgresql.yml
+++ b/.github/workflows/run_tests_postgresql.yml
@@ -41,4 +41,6 @@ jobs:
       run: python -m unittest discover -s cornflow.tests.unit
       env:
         FLASK_ENV: testing
+        POSTGRES_HOST: localhost
+        POSTGRES_PORT: 5432
         DATABASE_URL: postgres://postgres:postgresadmin@127.0.0.1:5432/cornflow-test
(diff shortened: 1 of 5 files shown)


.github/workflows/run_tests_postgresql.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -U -r requirements-dev.txt
+    - name: Run unit tests
+      run: python -m unittest discover -s cornflow.tests.unit
+      env:
+        FLASK_ENV: testing
+        POSTGRES_HOST: localhost
+        POSTGRES_PORT: 5432
```
pchtsp, line 45: should these two be here? or in the postgres service configuration above?

ggsdc: In the examples given by GitHub they place those values there:

https://docs.github.com/en/actions/guides/creating-postgresql-service-containers#running-jobs-directly-on-the-runner-machine

cornflow/tests/unit/test_roles.py
```diff
@@ -61,11 +61,11 @@ def test_get_no_roles(self):
                 )
                 self.assertEqual(403, response.status_code)
```
ggsdc, line 63: This is referenced in issue #77 , I think I already know why this is happening, but I want to try something out and leave the issue open with the idea of how to fix it so someone can do it.

.github/workflows/run_tests_postgresql.yml
pchtsp: I checked the example: actually we do not need those env vars.

Check what they say in the comments:
```
# Runs a script that creates a PostgreSQL table, populates
# the table with data, and then retrieves the data
run: node client.js
# Environment variables used by the `client.js` script to create
# a new PostgreSQL table.
env:
  # The hostname used to communicate with the PostgreSQL service container
  POSTGRES_HOST: localhost
	  # The default PostgreSQL port
	  POSTGRES_PORT: 5432
```
So the environment variables are for the node client.js example.

ggsdc: I have deleted the environment variables and it works. I will perform the merge.

PR #92, opened 2021-10-11
Docs

pchtsp:
I think I used an old `docs` branch instead of creating a new `docs` branch from master. So it looks like there's a lot of changes but in fact it's only one commit with a few changes.


pchtsp:
merge, please.

PR #93, opened 2021-10-19
Docs

pchtsp:


docs/source/guides/deploy_solver_new.rst
```diff
@@ -105,18 +128,75 @@ Schema property
 
 The schema of an application is the configuration used to solve a problem. This schema needs to have at least the `timeLimit` and the `solver` properties. Besides that, it's up to the developer to decide which configuration is needed. If you want to know how to define a json-schema, check the section :ref:`Write a json-schema`.
 
+Conventions
+*****************
+
+We follow some common conventions in the configuration schema so most apps share most of the main properties. Below is a list of known properties and what the usually represent.
+
+#. **timeLimit**: float. It indicates the amount of seconds before the method should stop.
```
ggsdc, line 136: This may be confusing, because when using a mathematical solver (cbc or cplex, for example) we use that as the time limit for the mathematical solver, not the solver itself, so the time it takes to solve the problem could be bigger based on the time needed to process the data and the solution.

I do not know if this merits a mention or not.

docs/source/guides/deploy_solver_new.rst
```diff
@@ -105,18 +128,75 @@ Schema property
...
 
+Each app is contained ideally inside a directory.
+
+The :ref:`Instance class` is implemented in the ``my-project/core/instance.py`` file, the :ref:`Solution class` inside the ``my-project/core/solution.py``.
+
+Each :ref:`Solution method` is defined in the ``my-project/solvers`` directory: ``solver_1.py``, ``solver_2.py``, etc.
+
+The :ref:`Application class` is defined inside ``my-project/__init__.py``.
+
+Schemas are stored in the ``my-project/schemas`` folder. Finally, :ref:`Test cases` are stored in the ``my-project/data`` directory.
```
ggsdc, line 193: I would change my-project to my_project because while importing, for example, in the test we can not do: `from DAG.my-project import Application`.

As it will throw an error because python modules can not use the `-` but it can use the underscore so `from DAG.my_project import Application` will work

docs/source/guides/deploy_solver_new.rst
pchtsp: if the pre-processing is long enough to do something, then whoever uses that should substract the time it takes to pre-process before passing the parameter to the mathematical solver. It could merit a mention but I'm not sure where. We do not mention mathematical solvers here so I don't think it makes sense to include the term.

docs/source/guides/jsonschema.rst
```diff
@@ -201,6 +201,7 @@ When naming columns in a "master table", we refer to the unique id of each row a
         "required": ["shifts", "resources_unavailable"]
     }
 
+As explained in the section beforehand, the parameters that are unidimensional should be on a table called ``parameters``.
 
 Example with TSP
 -------------------
```
ggsdc, line 207: I do not understand why it is complaining here. In the github viewer it can be seen properly and on pycharm the same

docs/source/guides/coding_style.rst
```diff
@@ -0,0 +1,96 @@
...
+        this.colorBase = '#65A0D1'
+        this.colorSource = '#D35A5A'
+
+        if (this.instance != null) {
+          this.horizon = this.instance.data.parameters.horizon
+          this.unit = this.instance.data.parameters.unit
+        }
+      }
+
+In the example, we're checking for the existance of instance and solution when creating a new Experiment.
```
ggsdc, line 66: From this we're onwards github seems to not get properly the syntax highlighting and is even raising an error on line 91

docs/source/guides/coding_style.rst
```diff
@@ -0,0 +1,96 @@
+Coding style and rules
+============================
+
+The following rules and notation come from examples in PR-revisions that we think can become common rules. Each rule will have an example and a description.
+
+
+Commenting out code
+-------------------------
+
+Having commented code makes no sense. It should be eliminated from the commit altogether. If for some reason you fear losing the piece of code in the future, remember that it will be present in the git history (if it has taken part of the master branch at some point):
```
ggsdc, line 10: I would make a little more emphasis on the if it has taken part of the master branch at some point, maybe having instead of having it between parenthesis, something like this:

"[...] git history, if it has taken part of the **master branch** at some point:"

docs/source/guides/coding_style.rst
```diff
@@ -0,0 +1,225 @@
...
+                time += self.instance.get_customer_property(
+                    current_stop, "setupTime"
+                )
+            total_cost += distance * dist_cost + time * time_cost
+
+
+    if layover:
+        total_cost += layover_cost
+
+In general, I do not like excessive indentation. One way to do that with an if inside a for is with something like this::
```
ggsdc, line 142: I would avoid to have I in the documentation. Rephrase to something like:

"In general, excessive indentation can make the code hard to review and understand. One way..."

PR #143, opened 2021-12-01
Quick start guide

marioncottard:
First version of a quick start guide


marioncottard:
I never used a local cornflow server so I am not a 100% sure about the part about launching it. I used the documentation to do that part, but I don't know if it up to date

PR #169, opened 2021-12-14
Hotfix/fernet key

ampuerin:
Fix environment variable about fernet key. Now we are using AIRFLOW__CORE__FERNET_KEY.


ggsdc (approved):
Just changed objective branch to release

Issue #80, opened 2021-07-28
Control dag-level permissions in cornflow / airflow


pchtsp:
**Is your feature request related to a problem? Please describe.**
Sometimes I'm a user who should not have access to certain solvers.

**Describe the solution you'd like**
I would like to be able to assign dag permissions to roles so only certain users would have access to certain dags/solvers.

**Describe alternatives you've considered**
Maybe we could store that in airflow and then cornflow would then use the cornflow user to get the dag information from airflow?
https://airflow.apache.org/docs/apache-airflow/stable/security/access-control.html#dag-level-role

ggsdc:
Couldn't we implement this with user group? And the group the user belongs
to, indicates the dags the user have access to, that way they can not even
create an instance pointing to that schema.

pchtsp:
It's possible and it's true we have talked about it.

Currently, cornflow does not have a list of avaliable dags. If we decide to
keep the dags only in airflow, then maybe it makes sense to control access
there. That way cornflow asks airflow (as it does now) for information on
dags. This implies we would need to replicate cornflows roles in airflow.

If we decide to create a dag catalog in cornflow that is synced with
airflow, then we could give permission at the dag level.
This implies that we would have to replicate airflow dag information in
cornflow.

There are probably other factors that I'm not thinking of that are
relevant.

ggsdc:
Should be closed with #182

ggsdc:
Cornflow-admin changes to be made on issue #198

ggsdc:
Solved, missing cornflow-admin update

PR #170, opened 2021-12-14
Control DAG level permissions on cornflow

ggsdc:
This pull request should cover issue #80 

There is some cleanup on the cornflow-server side as well.
Added rollbacks on the commands in case of integrity errors


ggsdc:
There is some changes still missing that will go on other branchs to the realease/v3.0 where I will accumulate everything, but you can start reviewing this PR @marioncottard @ampuerin

cornflow/commands/schemas.py
```diff
@@ -0,0 +1,34 @@
...
+        if verbose == 1:
+            print(f"Airflow is not reachable (attempt {attempts})")
+        time.sleep(15)
+
+    if not af_client.is_alive():
+        if verbose == 1:
+            print("Airflow is not reachable")
+        return False
+
+    response = af_client.update_schemas()
```
marioncottard, line 25: I can't find the `update_schemas()` function anywhere in the cornflow_client repository, is it because you haven't created it yet?

cornflow/tests/integration/test_cornflowclient.py
```diff
@@ -354,9 +354,13 @@ def test_manual_execution2(self):
...
         one_instance = self.create_new_instance("./cornflow/tests/data/test_mps.mps")
         payload = dict(
-            name="bla", config=dict(solver="CBC"), instance_id=one_instance["id"]
+            name="bla",
+            config=dict(solver="CBC"),
+            instance_id=one_instance["id"],
+            schema="solve_model_dag",
         )
         execution = self.client.create_api("execution/?run=0", json=payload)
+        print(execution.json())
```
marioncottard, line 363: This print needs to be removed

cornflow/commands/actions.py
```diff
@@ -16,7 +18,10 @@ def register_actions_command(verbose):
     if len(actions_to_register) > 0:
         db.session.bulk_save_objects(actions_to_register)
 
-    db.session.commit()
+    try:
+        db.session.commit()
+    except IntegrityError:
+        db.session.commit()
```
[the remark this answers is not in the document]
marioncottard: Same comments in `commands/roles.py`, `commands/permissions.py` and `commands/views.py`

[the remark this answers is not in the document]
ggsdc: This has been a bad copy and paste.... Will fix

cornflow/shared/authentication.py
```diff
@@ -120,6 +130,36 @@ def decorated_user(*args, **kwargs):
...
+    def dag_permission_required(func):
+        """
+        DAG permission decorator
+        :param func:
+        :return:
+        """
+
+        @wraps(func)
+        def dag_decorator(*args, **kwargs):
+            if int(current_app.config["OPEN_DEPLOYMENT"]) == 1:
```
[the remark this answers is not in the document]
ggsdc: You are right, this was part of my tests, forgot to change it back.

If it is an open deployment (1) then we do not restrict the access to the DAGs based on the user, so this decorator should do nothing, that is why only it works on one case

cornflow/commands/schemas.py
ggsdc: It is in a branch of cornflow_client (feature/list_dags)

cornflow/tests/integration/test_cornflowclient.py
ggsdc: Done

PR #171, opened 2021-12-15
Added endpoint for token validation

marioncottard:
Related to https://github.com/baobabsoluciones/cornflow-server/issues/142.

### DIFF
--- a/cornflow/endpoints/__init__.py
+++ b/cornflow/endpoints/__init__.py
@@ -26,6 +26,7 @@ from .execution import (
 )
 
 from .health import HealthEndpoint
+from .token import TokenEndpoint
 
 from .instance import (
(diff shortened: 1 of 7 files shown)


marioncottard:
I added the endpoint in the LoginEndpoint class because it was related, but I don't know if it is the best place to put it, what do you think? Also I was not sure about what unitary tests needed to be added and where.

ggsdc:
> I added the endpoint in the LoginEndpoint class because it was related, but I don't know if it is the best place to put it, what do you think? Also I was not sure about what unitary tests needed to be added and where.

I think it would be ebst to have this endpoint separate from the login endpoint, something like the health endpoint

marioncottard:
Done

PR #172, opened 2021-12-15
Password and email pattern validation

marioncottard:


marioncottard:
Can you help me @ggsdc ? I have no idea why the integration fails now

ggsdc:
There seems that there is something wrong with the solving of that case in particular, but I can not find why right now.

Im going to change the destination branch to the release/v3.0 branch and will add a TODO to reactivate the test again, but I need to do some further tests.

ggsdc (approved):

marioncottard:
Okay. It might be because of one of the passwords I had to change. I just noticed than in `.github/workflows/run_test_airflow.yml`, in the env variable `AIRFLOW_CONN_CF_URI`, there is the password 'airflow_test_password'. I know that some of the service users had the same password as this one but I had to change them because of the pattern validation. Could this have caused a problem in the communication with airflow, if this password and the ones in the service users are not the same?

ggsdc:
> Okay. It might be because of one of the passwords I had to change. I just noticed than in `.github/workflows/run_test_airflow.yml`, in the env variable `AIRFLOW_CONN_CF_URI`, there is the password 'airflow_test_password'. I know that some of the service users had the same password as this one but I had to change them because of the pattern validation. Could this have caused a problem in the communication with airflow, if this password and the ones in the service users are not the same?

I don't think that is the issue, as the user gets sing up properly and just one test is the one failing.

PR #178, opened 2021-12-20
Added endpoint for pwd recovery

marioncottard:
Linked to https://github.com/baobabsoluciones/cornflow-server/issues/164


marioncottard:
For now I created an email address for Cornflow-server, to run the tests and develop the endpoint, but maybe at some point it would be good to have an "official" one if there isn't one yet

ggsdc (changes requested):
The changes are good, but I think you should delete that gmail account as this is a public repo and you have the email and the password, I will be finishing this with David in order to have an email that we can use but that we do not save here the credentials or the email so that we do not have security problems

marioncottard:
Okay I just deleted it. I will comment the related unitary tests because otherwise they will fail

ggsdc:
@marioncottard can you review it?

ggsdc (approved):

ggsdc:
We are missing documentation, but I guess that that should be merged with some other parts that we are missing. And unit tests, which I do not know what to test, I have to think about it.

marioncottard:
Okay I'll review it now

cornflow/endpoints/user.py
```diff
@@ -174,3 +173,53 @@ def put(self, user_id, make_admin):
...
+            or email_config["email_receiver"] is None
+        ):
+            raise InvalidUsage(
+                "This functionality is not available. Check that cornflow's email is correctly configured"
+            )
+
+        message = f"A new password was generated for user with email: {email}"
+
+        if not UserModel.check_email_in_use(email):
+            return {"message": message}, 200
```
marioncottard, line 214: I don't think this is right: the condition checks if the email corresponds to an user in the database. If it doesn't, it should return an error, or at least a warning saying that the password could not be changed because the user wasn't found

ggsdc: The thing is I do not want to give information about which emails exist on the database and which don't. That is why the message back is the same. I think I'm going to modify it to make it more standard, but I do not want to make that distinction

marioncottard: Okay, that makes sense

Issue #134, opened 2021-11-10
Create unit test fot the cascade deletion of permissions


ggsdc:
Modify the current test `test_permission_cascade_deletion` in `test_roles` to be equal to `test_user_role_cascade_deletion`.

This has to be done after issue #133

PR #204, opened 2022-01-24
Added cascade test for the permissions.

ggsdc:
Modified migration that had a small error.
Solves issue #134

### DIFF
--- a/cornflow/tests/unit/test_permissions.py
+++ b/cornflow/tests/unit/test_permissions.py
@@ -11,9 +11,15 @@ from cornflow.endpoints import (
     PermissionsViewRoleEndpoint,
     PermissionsViewRoleDetailEndpoint,
 )
-from cornflow.models import InstanceModel, PermissionViewRoleModel
-from cornflow.shared.const import ROLES_MAP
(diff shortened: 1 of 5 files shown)


cornflow/tests/unit/test_permissions.py
```diff
@@ -209,3 +215,30 @@ def test_create_instance_no_permissions(self):
...
+        before_permissions = PermissionViewRoleModel.get_all_objects()
+        action = ActionModel.get_one_object(GET_ACTION)
+        action.delete()
+        after_permissions = PermissionViewRoleModel.get_all_objects()
+        self.assertNotEqual(before_permissions, after_permissions)
+
+    def test_permission_api_view_cascade_deletion(self):
+        before_permissions = PermissionViewRoleModel.get_all_objects()
+        api_view = ApiViewModel.get_one_by_name("instance")
+        print(api_view)
```
marioncottard, line 241: This print needs to be deleted

marioncottard (commented):
Just one thing, otherwise it looks good to me

cornflow/tests/unit/test_permissions.py
ggsdc: I always leave one or two, my bad

PR #208, opened 2022-01-27
Feature/gurobi optimizer installation

ampuerin:
Install Gurobi Optimizer into airflow service
New user and password policy on docker-compose example files


ggsdc:
Can you solve the conflicts? or do you need help?

ggsdc (approved):

Issue #191, opened 2022-01-18
Export check_data to instance table


ggsdc:
Extends issue #100 so that the result of the checks performed on the instance can be stored on the database and communicated to the user either as warnings or errors, so the user can fix the data errors.

Issue #190, opened 2022-01-18
Export check_solution to executions table


ggsdc:
The check_solution method creates a dictionary with different checks that have been run on the instance / solution after solving. Create a new field on the executions model so this dict can be stored as a json (after being created and passing a json schema validation) so that the results of these checks can be shown to the user either as warnings or errors in the solution.

PR #213, opened 2022-02-08
Feature/check solution save

ggsdc:
Added checks column to instance, execution and case table.
Solves #190
Solves #191


cornflow/endpoints/case.py
```diff
@@ -113,6 +113,7 @@ def get_execution_data(execution_id):
                 raise ObjectDoesNotExist("Execution does not exist")
             data = get_instance_data(execution.instance_id)
```
marioncottard, line 114: I'm not sure about this but shouldn't you update the `get_instance_data()` function so that it also returns the instance checks ?

ggsdc: You are right. Changed

marioncottard (dismissed):

PR #246, opened 2022-03-23
log_text EndPoint

magg2607:


ggsdc (approved):

ggsdc:
Iam going to merge the PR but you should add a unit tests to check that both logs are in the response body

Issue #259, opened 2022-03-31
Create process to automatically create REST API from json schema.


ggsdc:
This should be a new module (cornflow-tools maybe?) which helps you create the sqlalchemy models, base schemas and endpoints from a JSON Schema.

PR #265, opened 2022-04-07
Feature/cornflow tools

marioncottard:
Linked to [#259](https://github.com/baobabsoluciones/cornflow/issues/259)


ggsdc:
A couple of things:

1. I'm not going to merge it right away, cause maybe this should be on the library that I'm working on. cornflow-core, which you can see on branch feature/cornflow_core.
2. Could we try to have unit tests of this, to check that is working properly? Does it make sense.
3. If I understand correctly, this takes a json schema and generates a set of models / schemas / endpoints. How much would it take to given a set of models create a json schema?

marioncottard:
> A couple of things:
> I'm not going to merge it right away, cause maybe this should be on the library that I'm working on. cornflow-core, which you can see on branch feature/cornflow_core.

Okay!

> Could we try to have unit tests of this, to check that is working properly? Does it make sense.

Yes it makes sense. I'll add some. Just to be sure, the tests should only verify that the method generates all the classes needed, and that they have all the required properties/methods, right?

> If I understand correctly, this takes a json schema and generates a set of models / schemas / endpoints. How much would it take to given a set of models create a json schema?

Yes exactly. 
I think it is doable. I'll make it into another module. In that case, the arguments of the function would be the paths to the models' files, right ?

ggsdc:
> > A couple of things:
> > I'm not going to merge it right away, cause maybe this should be on the library that I'm working on. cornflow-core, which you can see on branch feature/cornflow_core.
> 
> Okay!
> 
> > Could we try to have unit tests of this, to check that is working properly? Does it make sense.
> 
> Yes it makes sense. I'll add some. Just to be sure, the tests should only verify that the method generates all the classes needed, and that they have all the required properties/methods, right?
> 

That should be enough, yes.

> > If I understand correctly, this takes a json schema and generates a set of models / schemas / endpoints. How much would it take to given a set of models create a json schema?
> 
> Yes exactly. I think it is doable. I'll make it into another module. In that case, the arguments of the function would be the paths to the models' files, right ?

The folder where the models are, just exclude any `__init__ `file in that folder.

ggsdc:
More information, you can use the `click` library to be able to run python code from the command line, so you could have something like:

> cornflow_tools generate_models "route_to_the_jsonschema"

And do it directly on the command line

marioncottard:
> More information, you can use the click library to be able to run python code from the command line, so you could have something like:

>> cornflow_tools generate_models "route_to_the_jsonschema"

> And do it directly on the command line

I'm not sure how this is different from what I did ? Is it because it is better to call directly the module instead of calling python first  in the command line (ex: `python generate_models "route_to_the_json_schema"`) ?

marioncottard:
@ggsdc I finished adding the second module and the unit tests. 

About that: 
> More information, you can use the click library to be able to run python code from the command line, so you could have something like:

>>cornflow_tools generate_models "route_to_the_jsonschema"

>And do it directly on the command line

Can you explain exactly what I should change because I don't get the difference between this and what I did with argparse

ggsdc:
> @ggsdc I finished adding the second module and the unit tests.
> 
> About that:
> 
> > More information, you can use the click library to be able to run python code from the command line, so you could have something like:
> 
> > > cornflow_tools generate_models "route_to_the_jsonschema"
> 
> > And do it directly on the command line
> 
> Can you explain exactly what I should change because I don't get the difference between this and what I did with argparse

Maybe you are right and there is no difference, we should just try it out what happens if we deploy conrflow-tools as a package on pypi or through github. I will give it a go after the holidays and we can review what are the missing changes needed to be done.

marioncottard:
There might be a few more changes to make since cornflow-core is not completely finished but I think it will only be smaller changes

ggsdc:
@marioncottard can you review it? I have moved things around and made some small changes to how the cli functions are implemented. I have to update the README and it should be done

marioncottard:
There was a change missing in the README, but now it looks good to me

ggsdc (approved):

PR #275, opened 2022-04-28
new volume path on docker-compose files

sistemas-baobab:
New volume path to mount example cornflow-dags folder into docker-airflow service


docker-compose.yml
```diff
@@ -37,9 +37,9 @@ services:
       dockerfile: Dockerfile
     restart: always
     volumes:
-      - ./airflow_config/dags:/usr/local/airflow/dags
-      - ./airflow_config/requirements.txt:/requirements.txt
-      - ./airflow_config/plugins:/usr/local/airflow/plugins
+      - ./cornflow-dags/DAG:/usr/local/airflow/dags
+      - ./cornflow-dags/requirements.txt:/requirements.txt
+      - ./cornflow-server/airflow_config/plugins:/usr/local/airflow/plugins
```
ggsdc, line 42: This last line, should not need to be changed as the context of the docker compose file is `./cornflow-server/airflow_config`, no?

ggsdc: This same thing could be said about the rest of the docker compose files

ggsdc (commented):
We should talk about this on tuesday.

docker-compose.yml
ampuerin: Now, the context of docker-compose files is in root repo folder. We have to load plugins from ./cornflow-server/airflow_config/plugins.
For DAG and requirements, now been moved to new DAG folder in the repo we load them from ./cornflow-dags/DAG and ./cornflow-dags/requirements.txt.

ggsdc (approved):

PR #276, opened 2022-04-28
Feature/kpi solutions

### LINKED ISSUE #251: Add key performance indicators on solutions
The data store on the execution should be able to have a indicators table (in the same way the instance usually has a parameters table).

This could allow to give back the indicators only when we query for cases, loading the indicators values for the solution in the data, making easier to get a quick view of how good or bad the case is compared with others.

JaimeSotomayor:
Closes #251

### DIFF
--- a/cornflow-server/cornflow/schemas/case.py
+++ b/cornflow-server/cornflow/schemas/case.py
@@ -4,6 +4,8 @@ and to serialize the response data given by the same endpoints.
 """
 
 # Imports from marshmallow library
+import json
+
 from marshmallow import fields, Schema
 
 # Import from internal modules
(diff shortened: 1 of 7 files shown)


cornflow-server/cornflow/schemas/case.py
```diff
@@ -4,6 +4,8 @@
 """
 
 # Imports from marshmallow library
+import json
+
```
ggsdc, line 8: This import does not get used. Delete it

cornflow-server/cornflow/tests/unit/test_cases.py
```diff
@@ -148,6 +148,7 @@ def test_new_case_execution(self):
         self.payload["solution"] = self.execution.data
         self.payload["solution_hash"] = self.execution.data_hash
         self.payload["user_id"] = self.user
+        self.payload["indicators"] = ""
 
         for key in self.response_items:
             self.assertEqual(self.payload[key], getattr(created_case, key))
```
ggsdc, line 154: As `indicators` is not in `self.response_items` adding it to the `self.payload` doesn't do anything as it is not getting checked against the response of the endpoint. You are not testing if the change you made is actually valid or not.

ggsdc (changes requested):
Fix the unit tests to really test if it is working.

cornflow-server/cornflow/tests/unit/test_cases.py
JaimeSotomayor: All done

PR #277, opened 2022-04-29
Feature/example dag data

### LINKED ISSUE #89: endpoint to return example data from a solution method
Not related.

In cornflow, apps come with their own test datasets (small) to test it works. We could have an endpoint in cornflow server to download sample data from a solution method. In order to do this, we would have to store the data somehow in airflow, probably (like the schemas). Or have a dag that automatically writes these datasets in the cornflow database.

This way, cornflow-app could load example data from cornflow-server. Also, it would be easier to have examples that don't need external data. Finally, they will give the potential user with an explicit working example of a dataset.

**Describe alternatives you've considered**
Alternatives: have the datasets in the cornflow-server github. But not ideal.

**Additional context**
No additional context.

JaimeSotomayor:
Closes #89


cornflow-server/cornflow/endpoints/example_data.py
```diff
@@ -0,0 +1,65 @@
...
+            if not af_client.is_alive():
+                log.error(
+                    "Airflow not accessible when getting schema {}".format(dag_name)
+                )
+                raise AirflowError(error="Airflow is not accessible")
+
+            # try airflow and see if dag_name exists
+            af_client.get_dag_info(dag_name)
+
+            log.debug("User gets exampple data from {}".format(dag_name))
```
ggsdc, line 52: exampple -> example

cornflow-server/cornflow/tests/unit/test_example_data.py
```diff
@@ -0,0 +1,57 @@
...
+            self.url + "{}/".format(self.schema_name),
+            {},
+            expected_status=200,
+            check_payload=False,
+        )
+        self.assertIn("examples", example)
+        self.assertIn("name", example)
+        self.assertEqual(example["examples"], self.example)
+        af_client.is_alive.assert_called_once()
+        af_client.get_dag_info.assert_called_once()
```
ggsdc, line 57: In this case this test should be on the integration folder

ggsdc (changes requested):
You need to make several changes in order for the PR to work and make sense.
**DO NOT COPY CODE WITHOUT UNDERSTANDING IT**

ggsdc:
One more thing. Once this PR is closed and the DAG is accepted, you should open an similar PR on cornflow-dags (the private repo) bringing this changes as well.

cornflow-dags/DAG/dag_timer.py
```diff
@@ -40,5 +40,6 @@ class Timer(ApplicationCore):
         solvers=list(solvers.keys()),
     )
 
-    def test_cases(self):
-        return []
+    @property
+    def test_cases(self) -> List[Dict]:
+        return list()
```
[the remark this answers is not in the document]
JaimeSotomayor: imported

ggsdc (approved):

PR #287, opened 2022-05-18
Added configuration to activate/deactivate signup

alejandro-bob:


cornflow-server/cornflow/config.py
```diff
@@ -18,6 +18,7 @@ class DefaultConfig(object):
     DEBUG = True
     TESTING = True
     LOG_LEVEL = int(os.getenv("LOG_LEVEL", 30))
+    SIGNUP_ACTIVATED = 1
```
ggsdc, line 21: Ok, change this line to:

```python 
SIGNUP_ACTIVATED = int(os.getenv("SIGNUP_ACTIVATED", 1))
```

cornflow-server/cornflow/app.py
```diff
@@ -28,7 +28,7 @@
     create_service_user_command,
 )
 from .commands.views import register_views_command
-from .config import app_config
+from .config import app_config, DefaultConfig
```
ggsdc, line 31: You do not need to import `DefaultConfig`

cornflow-server/cornflow/app.py
```diff
@@ -80,7 +80,8 @@ def _fk_pragma_on_connect(dbapi_con, con_record):
     auth_type = app.config["AUTH_TYPE"]
 
     if auth_type == AUTH_DB:
-        api.add_resource(SignUpEndpoint, "/signup/", endpoint="signup")
+        if DefaultConfig.SIGNUP_ACTIVATED:
+            api.add_resource(SignUpEndpoint, "/signup/", endpoint="signup")
```
ggsdc, line 84: Here do:

```python
signup_activated = int(app.config["SIGNUP_ACTIVATED"])
if signup_activated == 1:
    api.add_resource(SignupEndpoint, "/signup/", endpoint="signup")
```

ggsdc (changes requested):
Some changes to do, and you are missing adding test case to test that is working properly. To do this you can use as a base something like this:

``` python
class TestSignUpDeactivated(CustomTestCase):
    def create_app(self):
        app = create_app("testing")
        app.config["SIGNUP_ACTIVATED", 0]
        return app

    def test_signup_deactivated(self):
        ....
```
And in the test_signup_deactivated you try to perform a sign up a check that the code that gives you back is a 5XX I think.

ggsdc (dismissed):

PR #292, opened 2022-05-24
Feature/client queue

ggsdc:


ggsdc:
@alejandro-bob these are the main changes you made on cornflow client with some small fixes, review them and if you have any doubt about the changes I've made we can talk about them.

PR #302, opened 2022-07-01
Feature/check schemas

marioncottard:


ggsdc:
I've made some changes to the client library without merging this and now we have conflicts. @marioncottard could you review it and fix it so I can merge the changes.

cornflow-dags/requirements.txt
```diff
@@ -11,3 +11,5 @@ openpyxl
 pyomo
 acopy
 tsplib95
+simpy
+scipy
```
ggsdc, line 15: Why there is these two dependencies here? They are not used in any DAG. If they are a dependency of another DAG that is in another library, then they should be in that library requirements not here

cornflow-dags/requirements.txt
```diff
@@ -1,8 +1,8 @@
-cornflow-client
+cornflow-client @ git+https://github.com/baobabsoluciones/cornflow@feature/check_schemas#subdirectory=libs/client
```
ggsdc, line 1: I have removed this dependency like this.

ggsdc (approved):
@marioncottard Ok, everything is reviewed and seems alright. Great job!

Just a couple of things:
1. I have applied black formatting to all modified files, you have your formatter deactivated. Check it out.
2. I have change the way the update works on instances, executions and cases. Now they delete the checks and then call `super().update(data)` as it should be providing the same functionality as before.

Please review this last change just in case.

ggsdc (dismissed):

PR #323, opened 2022-10-06
correction of put_detail

hugolarzabal:
currently put_detail and patch_detail do not allow to edit if the user is different.
This PR, correct this issue.

### DIFF
--- a/libs/core/cornflow_core/resources/meta_resource.py
+++ b/libs/core/cornflow_core/resources/meta_resource.py
@@ -7,7 +7,7 @@ from pytups import SuperDict
 
 # Import from external libraries
 from flask_restful import Resource
-from flask import g, request
+from flask import g, request, current_app
 from flask_apispec.views import MethodResource
 
 # Import from internal modules
@@ -31,6 +31,7 @@ class BaseMetaResource(Resource, MethodResource):
         self.auth_class = None
         self.dependents = None
         self.unique = None
+        self.user_access_all_objects = current_app.config["USER_ACCESS_ALL_OBJECTS"]
         pass
 
     """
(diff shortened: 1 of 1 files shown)


ggsdc (changes requested):
Shouldn't we also change it for delete_detail? I guess the query it does can have the same mistake as well, right?

There is a config parameter about if a user can access another users objects. I think it would be better if based on the value of said configuration the user is poped from the kwargs or it stays, so that it gets passed to the query behind or not.

hugolarzabal:
currently, the delete_detail method doesn't check the user at all. Should it do it?

Also, I think there should be a different check and error message if the user is not allowed to access this object instead of leaving user in the kwargs. The current message "The data entity does not exist on the database" is very confusing when you know that the data exist.

ggsdc:
> currently, the delete_detail method doesn't check the user at all. Should it do it?
> 
> Also, I think there should be a different check and error message if the user is not allowed to access this object instead of leaving user in the kwargs. The current message "The data entity does not exist on the database" is very confusing when you know that the data exist.

The user should still be going in the kwargs as the query that is performed need to take it in the case that a use can only access it own objects. If all users can access all objects, then the user is taken out of the kwargs as the query does not needed.

The messages where that way on purpose. In case a user is trying to access information that is not his/hers we do not want to tell them if the information does exists but the user does not have access to it, I prefer to just tell them that it does not exist.

I will make the needed changes on Monday and create the appropriate tests to make sure that this works.

hugolarzabal:
If you do not want users to access data from other users, I think the get_detail method should also be changed.

Also, I would suggest putthing "the data doesn't exist or you do not have permission" in the message if you want to maintain it the same for both cases.

ggsdc:
> If you do not want users to access data from other users, I think the get_detail method should also be changed.
> 
> Also, I would suggest putthing "the data doesn't exist or you do not have permission" in the message if you want to maintain it the same for both cases.

Actually, no changes are needed on the endpoints to adapt it, just creating a get_one_object query on the meta model that all models inherit from is enough.

It is how it is working on the instances or executions models for example. I'm going to test this out and if it works then I will close this PR

PR #333, opened 2022-11-21
Added pyomo missing status codes

marioncottard:
- Added missing pyomos termination condition status. Those do not appear in pyomo documentation but they can be returned.
- When pyomo returns a warning, it means that there was no error but the problem could not be solved. See [here](https://github.com/Pyomo/pyomo/blob/d143fb484f1abbbdd7f156d60a49b13524244c5f/pyomo/opt/results/solver.py#L86): pyomo returns a warning when no solution is found or when the problem is infeasible. So the associated cornflow solution status should be SOLUTION_STATUS_INFEASIBLE

### DIFF
--- a/cornflow-dags/DAG/rostering/schemas/solution.json
+++ b/cornflow-dags/DAG/rostering/schemas/solution.json
@@ -37,7 +37,7 @@
       "description": "Indicators of solution",
       "show": true,
       "properties": {
-        "fo": {
+        "of": {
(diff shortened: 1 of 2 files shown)


marioncottard:
Also: corrected the solution schema of rostering because the method Experiment.get_indicators() defines the field "of", not "fo"

ggsdc (approved):

PR #339, opened 2022-12-21
Feature/ldap service

ggsdc:
LDAP now can have a different user base for the service user if needed


ampuerin:
We need to edit the[ docker-compose-cornflow-ldap.yml ](https://github.com/baobabsoluciones/cornflow/blob/master/docker-compose-cornflow-ldap.yml)file in order to append env vars in cornflow service:

```
 - LDAP_SERVICE_BASE: ou=service,dc=example,dc=org
 - CORNFLOW_SERVICE_USER: cornflow
```

ampuerin (dismissed):

Issue #335, opened 2022-11-25
Make config schema be validated by reading the schema deployed on Airflow


ggsdc:
Currently the config field of an execution gets validated on arrival to cornflow, This configuration does not have all the parameters that the solvers use, so it should be better if the config schema gets read from airflow, the same way it is done with the data for the instance.

PR #340, opened 2022-12-28
Feature/change validation

ggsdc:
The validation of the config and instance data when creating a execution now it is performed against jsonchema not against marshamallow objects.

This caused some validation errors to slip and also should solve the problem with the config validation in the issue #335


Lucasct77 (approved):

ggsdc:
@marioncottard this is the PR I talked you about

PR #357, opened 2023-01-25
mapping to translate parameters dempending on solver and modeler

AlejandroMus:


libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,89 @@
...
+    ('pump_passes', 'pyomo', 'gurobi') : 'PumpPasses' ,
+    ('heuristics', 'pyomo', 'gurobi') : 'Heuristics' ,
+    ('presolve', 'pyomo', 'gurobi') : 'Presolve' ,
+    ('threads', 'pyomo', 'gurobi') : 'threads' ,
+}
+
+PULP_GUROBI_TRANSLATING_MAPPING = {
+    ('is_mip', 'pulp', 'cbc') : 'mip' ,
+    ('time_limit', 'pulp', 'cbc') : 'timeLimit' ,
+}
```
ggsdc, line 136: Should this be gurobi?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,89 @@
...
+    ('lp_iteration_limit', 'pyomo', 'cbc') : 'lp/iterlim' ,
+    ('mem_limit', 'pyomo', 'cbc') : 'limits/memory' ,
+    ('solution_limit', 'pyomo', 'cbc') : 'limits/maxsol' ,
+    ('time_limit', 'pyomo', 'cbc') : 'limits/time' ,
+    ('primal-dual_gap', 'pyomo', 'cbc') : 'limits/gap' ,
+    ('lp_optimality_tol', 'pyomo', 'cbc') : 'numerics/lpfeastol' ,
+    ('pump_passes', 'pyomo', 'cbc') : 'heuristics/feaspump/maxdepth' ,
+    ('presolve', 'pyomo', 'cbc') : 'presolve' ,
+    ('threads', 'pyomo', 'cbc') : 'threads' ,
+}
```
ggsdc, line 150: Here it should be scip

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,89 @@
...
+
+PULP_SCIP_TRANSLATING_MAPPING = {
+    ('is_mip', 'pulp', 'cbc') : 'mip' ,
+    ('optimality_tol', 'pulp', 'cbc') : 'gapAbs' ,
+    ('time_limit', 'pulp', 'cbc') : 'timeLimit' ,
+    ('threads', 'pulp', 'cbc') : 'threads' ,
+    ('max_nodes', 'pyomo', 'cbc') : 'maxNodes' ,
+}
```
ggsdc, line 158: SCIP and PULP

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,89 @@
...
+
+PYOMO_HIGHS_TRANSLATING_MAPPING = {
+    ('presolve', 'pyomo', 'cbc') : 'presolve' ,
+    ('parallel', 'pyomo', 'cbc') : 'parallel' ,
+    ('crossover', 'pyomo', 'cbc') : 'run_crossover' ,
+    ('time_limit', 'pyomo', 'cbc') : 'time_limit' ,
+}
```
ggsdc, line 165: highs

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,89 @@
...
+
+PULP_HIGHS_TRANSLATING_MAPPING = {
+    ('is_mip', 'pulp', 'cbc') : 'mip' ,
+    ('optimality_tol', 'pulp', 'cbc') : 'gapAbs' ,
+    ('time_limit', 'pulp', 'cbc') : 'timeLimit' ,
+    ('threads', 'pulp', 'cbc') : 'threads' ,
+}
```
ggsdc, line 172: highs

ggsdc (changes requested):
If there is no difference between the name in pulp and pyomo, maybe we do not need that distinction. 

Could we have everything in the same dict? instead of having several dicts? that way when is going to get used is easier to call just one dict instead of having to call several

Can you make a change to one dag so that it uses this translation so we know that it works?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,89 @@
...
+    ('heuristics', 'pyomo', 'cbc') : 'heur' ,
+}
+
+PULP_CBC_TRANSLATING_MAPPING = {
+    ('is_mip', 'pulp', 'cbc') : 'mip' ,
+    ('optimality_tol', 'pulp', 'cbc') : 'gapAbs' ,
+    ('cutoff', 'pulp', 'cbc') : 'cuts' ,
+    ('time_limit', 'pulp', 'cbc') : 'timeLimit' ,
+    ('threads', 'pulp', 'cbc') : 'threads' ,
+    ('presolve', 'pyomo', 'cbc') : 'presolve' ,
```
ggsdc, line 112: There is a mistake in this line still.

ggsdc (changes requested):

ggsdc:
> If there is no difference between the name in pulp and pyomo, maybe we do not need that distinction.
> 
> Could we have everything in the same dict? instead of having several dicts? that way when is going to get used is easier to call just one dict instead of having to call several
> 
> Can you make a change to one dag so that it uses this translation so we know that it works?

@AlejandroMus what about this?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,74 @@
...
+    ('optimality_tol', 'pyomo', 'cbc') : 'dualT' ,
+    ('pump_passes', 'pyomo', 'cbc') : 'pumpC' ,
+    ('heuristics', 'pyomo', 'cbc') : 'heur' ,
+
+    ('is_mip', 'pulp', 'cbc') : 'mip' ,
+    ('optimality_tol', 'pulp', 'cbc') : 'gapAbs' ,
+    ('cutoff', 'pulp', 'cbc') : 'cuts' ,
+    ('time_limit', 'pulp', 'cbc') : 'timeLimit' ,
+    ('threads', 'pulp', 'cbc') : 'threads' ,
+    ('presolve', 'pulp', 'cbc') : 'presolve' ,
```
ggsdc, line 110: How do you set the gap in pyomo? I do not see the option. Are you sure that you have implemented all the needed parameters?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,74 @@
...
+    ('mip_gap', 'pyomo', 'gurobi') : 'MIPGap' ,
+    ('optimality_tol', 'pyomo', 'gurobi') : 'OptimalityTol' ,
+    ('branch_dir', 'pyomo', 'gurobi') : 'BranchDir' ,
+    ('pump_passes', 'pyomo', 'gurobi') : 'PumpPasses' ,
+    ('heuristics', 'pyomo', 'gurobi') : 'Heuristics' ,
+    ('presolve', 'pyomo', 'gurobi') : 'Presolve' ,
+    ('threads', 'pyomo', 'gurobi') : 'threads' ,
+
+    ('is_mip', 'pulp', 'gurobi') : 'mip' ,
+    ('time_limit', 'pulp', 'gurobi') : 'timeLimit' ,
```
ggsdc, line 128: Same here, how do you set up the gap for gurobi?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,74 @@
...
+    ('lp_optimality_tol', 'pyomo', 'scip') : 'numerics/lpfeastol' ,
+    ('pump_passes', 'pyomo', 'scip') : 'heuristics/feaspump/maxdepth' ,
+    ('presolve', 'pyomo', 'scip') : 'presolve' ,
+    ('threads', 'pyomo', 'scip') : 'threads' ,
+
+    ('is_mip', 'pulp', 'scip') : 'mip' ,
+    ('optimality_tol', 'pulp', 'scip') : 'gapAbs' ,
+    ('time_limit', 'pulp', 'scip') : 'timeLimit' ,
+    ('threads', 'pulp', 'scip') : 'threads' ,
+    ('max_nodes', 'pulp', 'scip') : 'maxNodes' ,
```
ggsdc, line 146: How do you set up the `optimality_tol` in pyomo?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,74 @@
...
+
+    ('presolve', 'pyomo', 'highs') : 'presolve' ,
+    ('parallel', 'pyomo', 'highs') : 'parallel' ,
+    ('crossover', 'pyomo', 'highs') : 'run_crossover' ,
+    ('time_limit', 'pyomo', 'highs') : 'time_limit' ,
+
+    ('is_mip', 'pulp', 'highs') : 'mip' ,
+    ('optimality_tol', 'pulp', 'highs') : 'gapAbs' ,
+    ('time_limit', 'pulp', 'highs') : 'timeLimit' ,
+    ('threads', 'pulp', 'highs') : 'threads' ,
```
ggsdc, line 156: Same here

ggsdc (changes requested):

libs/client/cornflow_client/constants.py
AlejandroMus: In this two lines, the gap is set
    ('mip_gap', 'pyomo', 'gurobi') : 'MIPGap' ,
    ('optimality_tol', 'pyomo', 'gurobi') : 'OptimalityTol' ,
These are most used options

libs/client/cornflow_client/constants.py
AlejandroMus: It seems that highs do not offer this option. The allowed parameters are (from their github webpage)

  --model_file arg        File of model to solve.
  --presolve arg          Presolve: "choose" by default - "on"/"off" are alternatives.
  --solver arg            Solver: "choose" by default - "simplex"/"ipm" are alternatives.
  --parallel arg          Parallel solve: "choose" by default - "on"/"off" are alternatives.
  --run_crossover arg     Run crossover after IPM: "on" by default - "choose"/"off" are alternatives.
  --time_limit arg        Run time limit (seconds - double).
  --options_file arg      File containing HiGHS options.
  --solution_file arg     File for writing out model solution.
  --write_model_file arg  File for writing out model.
  --random_seed arg       Seed to initialize random number generation.
  --ranging arg           Report cost, bound, RHS and basic solution ranging in any solution file: "off" by default - "on" is alternatives.
  --read_solution_file    File of solution to be read
  
  --version               Print version number

ggsdc: But in pulp interface with HiGHS it seems like they set up a `gapRel`: https://github.com/coin-or/pulp/blob/006753076be18ee857339d333a5db11e599f21a2/pulp/apis/highs_api.py#L274

ggsdc: And in pyomo it is set up as well: https://github.com/Pyomo/pyomo/blob/98697d529370dbf3e3ed367bb8fd3e1fa94c0077/pyomo/contrib/appsi/solvers/highs.py#L197

libs/client/cornflow_client/constants.py
ggsdc: This is still unresolved. With cbc and pyomo I do not see the `gapAbs` parameter

libs/client/cornflow_client/constants.py
ggsdc: And in pulp?

libs/client/cornflow_client/constants.py
ggsdc: But the name should be the same that the one used in pulp, no? Why it is `optimality_tol` in one and `lp_optimality_tol` in the other?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,84 @@
...
+    ('cutoff', 'pyomo', 'scip') : 'heuristics/subnlp/setcutoff' ,
+    ('nlp_iteration_limit', 'pyomo', 'scip') : 'heuristics/subnlp/itermin' ,
+    ('gap', 'pyomo', 'scip') : 'limits/gap ' ,
+    ('optimality_tol', 'pyomo', 'scip') : 'limits/absgap ' ,
+
+    ('is_mip', 'pulp', 'scip') : 'mip' ,
+    ('optimality_tol', 'pulp', 'scip') : 'gapAbs' ,
+    ('time_limit', 'pulp', 'scip') : 'timeLimit' ,
+    ('threads', 'pulp', 'scip') : 'threads' ,
+    ('max_nodes', 'pulp', 'scip') : 'maxNodes' ,
```
ggsdc, line 152: I do not understand why in pyomo the gap is set up with `primal-dual_gap` but in pyomo is `optimality_tol`.

ggsdc (commented):
I think that there is something that we are not on the same page.

libs/client/cornflow_client/constants.py
AlejandroMus: One is a relative gap, while the other one is absolute. Please, note that in both cases, optimality_tol points to gapAbs. 
In pulp does not exist the relative gap option. I may agree that the primal-dual_gap have to be changed to rel_optimality_tool (for consistence with Highs), but reference names to absolute and relative should be different

libs/client/cornflow_client/constants.py
AlejandroMus: This is outdated. The most updated version is where you commented about optimality_tol and primal-dual_gap.

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,84 @@
...
+    ('optimality_tol', 'pyomo', 'gurobi') : 'OptimalityTol' ,
+    ('feasibility_tol', 'pyomo', 'gurobi') : 'FeasibilityTol' ,
+    ('branch_dir', 'pyomo', 'gurobi') : 'BranchDir' ,
+    ('pump_passes', 'pyomo', 'gurobi') : 'PumpPasses' ,
+    ('heuristics', 'pyomo', 'gurobi') : 'Heuristics' ,
+    ('presolve', 'pyomo', 'gurobi') : 'Presolve' ,
+    ('threads', 'pyomo', 'gurobi') : 'threads' ,
+    
+    ('is_mip', 'pulp', 'gurobi') : 'mip' ,
+    ('time_limit', 'pulp', 'gurobi') : 'timeLimit' ,
```
ggsdc, line 129: Can pulp only set up this two parameters when using gurobi?

libs/client/cornflow_client/constants.py
AlejandroMus: Because gapAbs parameter is defined from a SuperClass of CBC_solver in pulp called LP_solver. That means, gapAbs is a name that cbc decided to give to such a parameter but *does not exists* in cbc per se.
There are several option that can be discussed to set the optimality_tol. Among them, I chose dualT.

libs/client/cornflow_client/constants.py
AlejandroMus: Unfortunately, the "direct" gurobi parameters accepted in pulp are: 

mip ([bool]) – if False, assume LP even if integer variables

msg ([bool]) – if False, no log is shown

options ([list]) –

timeLimit ([float]) – maximum time for solver (in seconds)

args –

kwargs – optional named options to pass to each solver, e.g. gapRel=0.1, gapAbs=10, logPath=””, 

Therefore, the only feasible option I see is to adapt the kwargs parameter to include in gapAbs, but should be done carefully

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,93 @@
...
+    ('iteration_limit', 'pulp', 'gurobi') : 'IterationLimit' ,
+
+
+    ('bar_tol', 'pyomo', 'scip') : 'numerics/barrierconvtol' ,
+    ('cutoff_breaker', 'pyomo', 'scip') : 'heuristics/shiftandpropagate/cutoffbreaker' ,
+    ('lp_iteration_limit', 'pyomo', 'scip') : 'lp/iterlim' ,
+    ('mem_limit', 'pyomo', 'scip') : 'limits/memory' ,
+    ('solution_limit', 'pyomo', 'scip') : 'limits/maxsol' ,
+    ('time_limit', 'pyomo', 'scip') : 'limits/time' ,
+    ('rel_optimality_tol', 'pyomo', 'scip') : 'limits/gap' ,
```
marioncottard, line 144: The parameter that you call `rel_optimality_tol` for scip and highs is the same as `mip_gap` for cbc and gurobi, right ? If so, they should have the same name.

marioncottard (commented):
Can you please make a change to one dag so that it uses this translation so we know that it works? (Preferibly a DAG that uses pulp, please, because this seems to be a little bit more adapted to Pyomo) I know Guille commented that before but I see you haven't done it so I figured you might have forgotten

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,93 @@
...
+    ('presolve', 'pulp', 'gurobi') : 'presolve' ,
+    ('mip_gap', 'pulp', 'gurobi') : 'MIPGap' ,
+    ('optimality_tol', 'pulp', 'gurobi') : 'gapAbs' ,
+    ('feasibility_tol', 'pulp', 'gurobi') : 'FeasibilityTol' ,
+    ('iteration_limit', 'pulp', 'gurobi') : 'IterationLimit' ,
+
+
+    ('bar_tol', 'pyomo', 'scip') : 'numerics/barrierconvtol' ,
+    ('cutoff_breaker', 'pyomo', 'scip') : 'heuristics/shiftandpropagate/cutoffbreaker' ,
+    ('lp_iteration_limit', 'pyomo', 'scip') : 'lp/iterlim' ,
```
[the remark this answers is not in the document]
AlejandroMus: Indeed. However, while for cbc and gurobi this parameter is a stopping criterion for any case, scip has it for lp and NLP problems. So, I created lp_iteration_limit and nlp_iteration_limit for the two allowed cases in this solver

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,93 @@
...
+    ('presolve', 'pyomo', 'gurobi') : 'Presolve' ,
+    ('threads', 'pyomo', 'gurobi') : 'threads' ,
+    
+    ('is_mip', 'pulp', 'gurobi') : 'mip' ,
+    ('time_limit', 'pulp', 'gurobi') : 'timeLimit' ,
+    ('presolve', 'pulp', 'gurobi') : 'presolve' ,
+    ('mip_gap', 'pulp', 'gurobi') : 'MIPGap' ,
+    ('optimality_tol', 'pulp', 'gurobi') : 'gapAbs' ,
+    ('feasibility_tol', 'pulp', 'gurobi') : 'FeasibilityTol' ,
+    ('iteration_limit', 'pulp', 'gurobi') : 'IterationLimit' ,
```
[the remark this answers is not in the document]
AlejandroMus: It is not clear to me it allows more parameters. Let me answer you first the question
_, why is it that ('optimality_tol', 'pulp', 'gurobi') : 'gapAbs' and ('optimality_tol', 'pyomo', 'gurobi') : 'OptimalityTol' ?_
pyomo and pulp have different api. Therefore, same prameters can (and do) have diffferent names. Regarding your point of the different number of passed parameters, actually, pulp does not accept the same parameters as pyomo. Pulp has a keyword "option" that accpets a list of options (which is different from the **kwargs parameter discussed in some answers above). Looking into the pulp api, it seems that this argument option is not  accepting a lot of new parameters (https://github.com/coin-or/pulp/blob/master/pulp/apis/coin_api.py line 23)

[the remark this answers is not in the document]
hugolarzabal: Sorry for intruding the conversation. I am receiving mails about this PR (for being subscribed to cornflow) and I was intrigued by this.
The easier way to know what pulp is doing with the parameters is looking at its [code](https://github.com/coin-or/pulp/blob/master/pulp/apis/gurobi_api.py).

pulp arguments seems to be a little inconsistents depending on the solvers and the interface (python or cmd lines) but for gurobi, gapAbs does not seem to be mapped to OptimalityTol:
```
             logPath="LogFile",
            gapRel="MIPGap",
            gapAbs="MIPGapAbs",
            threads="Threads",
```
The former is in the cmd line class. The python class does not seems to have a gapAbs argument.
However the python Gurobi class has a **kwargs argument: `**solverParams`
So it would need to be tested but I think you could actually map pukp exactly like pyomo for Gurobi.

As I said, pulp is not perfectly consistent and the cbc class does not seems to have a kwarg.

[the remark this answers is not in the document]
hugolarzabal: Also, there seems to be a confusion here between absolute gap and optimality tolerance for gurobi which are not the same parameter:
[OptimalityTol](https://www.gurobi.com/documentation/current/refman/optimalitytol.html#parameter:OptimalityTol)
[MIPGapAbs](https://www.gurobi.com/documentation/current/refman/mipgapabs.html#parameter:MIPGapAbs)

Personally, I have never used OptimalityTol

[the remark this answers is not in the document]
AlejandroMus: You are more than welcome, Hugo! Indeed, you are right and we discussed this issue in an internal call. In this link https://github.com/coin-or/pulp/blob/master/pulp/apis/gurobi_api.py (line 336) it is the map I wrote (CMD-based). I took quite a few of these arguments from the **kwargs, and I assumed that they are gurobi-liked spelled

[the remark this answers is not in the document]
hugolarzabal: Ah yes, if I am not mistaken the difference between python and cmd interface is that you need gurobipy installed for python interface, right?
So yes, you need to use the cmd interface and it does not have a kwarg.

libs/client/cornflow_client/constants.py
AlejandroMus: Not exactly, although I can change the names if desired. MIPGap is the relative gap in a MIP problem. Since gurobi is more generic, I wanted to specify that this gap is noly for MIP problems. I don't know how exact/formal would be to assume that the relative gap = MIP gap

libs/client/cornflow_client/constants.py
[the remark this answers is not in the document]
hugolarzabal: In any case, it might be a good idea to raise an issue in pulp in order to ask for a kwargs in all solver options and to ask them to review the consistency of arguments between interfaces.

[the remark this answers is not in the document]
AlejandroMus: Thanks for pointing out the confusion on the MIPGapAbs and OptimalityTol. It will be corrected in the new commit

libs/client/cornflow_client/constants.py
hugolarzabal: I would suggest using abs_gap and rel_gap for every solver instead of mip_gap.
You should not use optimality_tol for gaps, it is too long and can refer to something else (for gurobi for example)

libs/client/cornflow_client/constants.py
```diff
@@ -118,9 +118,7 @@
     ('mem_limit', 'pyomo', 'gurobi') : 'MemLimit' ,
     ('solution_limit', 'pyomo', 'gurobi') : 'SolutionLimit' ,
     ('time_limit', 'pyomo', 'gurobi') : 'TimeLimit' ,
-    ('mip_gap', 'pyomo', 'gurobi') : 'MIPGap' ,
-    ('optimality_tol', 'pyomo', 'gurobi') : 'OptimalityTol' ,
-    ('feasibility_tol', 'pyomo', 'gurobi') : 'FeasibilityTol' ,
+    ('abs_gap', 'pyomo', 'gurobi') : 'MIPGap' ,
```
hugolarzabal, line 121: MIPGap is relative gap, absolute gap is MIPGapAbs

libs/client/cornflow_client/constants.py
```diff
@@ -97,7 +97,7 @@
     ('iteration_limit', 'pyomo', 'cbc') : 'maxIt' ,
     ('solution_limit', 'pyomo', 'cbc') : 'maxSaved' ,
     ('time_limit', 'pyomo', 'cbc') : 'sec' ,
-    ('mip_gap', 'pyomo', 'cbc') : 'allow' ,
+    ('abs_gap', 'pyomo', 'cbc') : 'allow' ,
     ('optimality_tol', 'pyomo', 'cbc') : 'dualT' ,
```
hugolarzabal, line 101: for cbc, relative gap is ratio (i do not see it)

hugolarzabal (changes requested):

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,91 @@
...
+    ('heuristics', 'pyomo', 'gurobi') : 'Heuristics' ,
+    ('presolve', 'pyomo', 'gurobi') : 'Presolve' ,
+    ('threads', 'pyomo', 'gurobi') : 'threads' ,
+    
+    ('is_mip', 'pulp', 'gurobi') : 'mip' ,
+    ('time_limit', 'pulp', 'gurobi') : 'timeLimit' ,
+    ('presolve', 'pulp', 'gurobi') : 'presolve' ,
+    ('abs_gap', 'pulp', 'gurobi') : 'gapAbs' ,
+    ('feasibility_tol', 'pulp', 'gurobi') : 'FeasibilityTol' ,
+    ('iteration_limit', 'pulp', 'gurobi') : 'IterationLimit' ,
```
marioncottard, line 134: Could we have a `rel_gap` option for (pulp, gurobi)?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,91 @@
...
+    ('max_nodes', 'pulp', 'scip') : 'maxNodes' ,
+    ('lp_iteration_limit', 'pulp', 'scip') : 'lp/iterlim' ,
+    ('mem_limit', 'pulp', 'scip') : 'limits/memory' ,
+
+
+    ('presolve', 'pyomo', 'highs') : 'presolve' ,
+    ('parallel', 'pyomo', 'highs') : 'parallel' ,
+    ('crossover', 'pyomo', 'highs') : 'run_crossover' ,
+    ('time_limit', 'pyomo', 'highs') : 'time_limit' ,
+    ('rel_gap', 'pyomo', 'highs') : 'mip_rel_gap' ,
```
marioncottard, line 166: Could we have the absolute gap for (pyomo, highs) ?

libs/client/cornflow_client/constants.py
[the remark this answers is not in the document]
marioncottard: What about for the people that do have gurobipy installed and are using the python interface?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,91 @@
...
+
```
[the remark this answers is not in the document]
ggsdc: You are missing the link

libs/client/cornflow_client/constants.py
AlejandroMus: Yes, and I will include it. However, it only works for the CMD version. Should we make a clear statement?

libs/client/cornflow_client/constants.py
```diff
@@ -90,6 +90,91 @@
...
+    ('is_mip', 'pulp', 'highs') : 'mip' ,
+    ('abs_gap', 'pulp', 'highs') : 'gapAbs' ,
+    ('time_limit', 'pulp', 'highs') : 'timeLimit' ,
+    ('threads', 'pulp', 'highs') : 'threads' ,
+    ('rel_gap', 'pulp', 'highs') : 'gapRel' ,
+
```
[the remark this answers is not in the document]
AlejandroMus: Thanks for the comments. I've added gapAbs. I found this link https://www.gams.com/latest/docs/S_HIGHS.html 
where appear more parameters. Not sure how useful could be to add more...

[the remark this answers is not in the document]
hugolarzabal: I think at least heuristics can be interesting to add: `mip_heuristic_effort`

libs/client/cornflow_client/constants.py
```diff
@@ -107,6 +107,101 @@
...
+    ('time_limit', 'pulp', 'scip') : 'timeLimit' ,
+    ('threads', 'pulp', 'scip') : 'threads' ,
+    ('max_nodes', 'pulp', 'scip') : 'maxNodes' ,
+    ('iteration_limit', 'pulp', 'scip') : 'lp/iterlim' ,
+    ('mem_limit', 'pulp', 'scip') : 'limits/memory' ,
+    ('cutoff_breaker', 'pyomo', 'scip') : 'heuristics/shiftandpropagate/cutoffbreaker' ,
+    ('iteration_limit', 'pyomo', 'scip') : 'lp/iterlim' ,
+    ('presolve', 'pyomo', 'scip') : 'presolve' ,
+    ('solution_limit', 'pyomo', 'scip') : 'limits/maxsol' ,
+    ('nlp_iteration_limit', 'pyomo', 'scip') : 'heuristics/subnlp/itermin' ,
```
hugolarzabal, line 182: here there seems to be pyomo and pulp commands mixed together.
also iteration_limit is repeated.
I would advise putting the arguments always in the same order. That way, it will be easier to find them.

libs/client/cornflow_client/constants.py
```diff
@@ -107,6 +107,101 @@
     pl.LpStatusUndefined: STATUS_UNDEFINED,
 }
```
hugolarzabal, line 109: it may be nice to add a readme with all cornflow parameters names and definitions

hugolarzabal (changes requested):

libs/client/cornflow_client/constants.py
ggsdc: That will be done latter, I do not want to keep this more time open.

libs/client/cornflow_client/constants.py
ggsdc: Is this the last change? @AlejandroMus when do you think it will be done?

AlejandroMus: Just done the push a few min ago. It is passing the unit tests

hugolarzabal (dismissed):

marioncottard (approved):

ggsdc (approved):

ggsdc (approved):

PR #359, opened 2023-01-26
Feature/sprint6 t7 t9

marioncottard:


ggsdc (dismissed):
Everything seems in order. Are we sure that with the change nothing that we have currently deployed is going to broke?

marioncottard:
There are some methods from the old Cornflow class that are no longer accessible from Cornflow and are now only accessible from the raw class, but they are only  "internal" methods, like `api_for_id()`, `get_api_for_id()`, `create_api()`, etc. If you have some scripts that use them, then you would need to update them. Otherwise, no, nothing should be broken

ggsdc (approved):

PR #363, opened 2023-02-01
Unified status between dags

marioncottard:


ggsdc (approved):
Everything seems alright. Only thing that worries me is the problems with the coverage

ggsdc (approved):

marioncottard:
I can review the coverage part more in detail if you want. I think the problem was already there before but I can try and add tests to increase coverage.

ggsdc:
> I can review the coverage part more in detail if you want. I think the problem was already there before but I can try and add tests to increase coverage.

Do not worry about it now. Seems like the base coverage was wrong from an earlier test

PR #365, opened 2023-02-03
Feature/sprint6 t11

marioncottard:


ggsdc:
Ok, so the test do not pass on cornflow dags due to the use of the dag_utilities. It can not use the ti argument as it does not exist.

marioncottard:
Fixed

ggsdc (dismissed):

ggsdc (approved):

PR #367, opened 2023-02-09
Feature/init access permissions

AlejandroMus:


cornflow-server/cornflow/commands/permissions.py
```diff
@@ -55,6 +55,21 @@ def register_base_permissions_command(verbose):
...
+            permissions_registered.append(to_reg_permission)
+
+    tb_removed = []
+    for i, reg_permission in enumerate(permissions_registered):
+        if reg_permission not in permissions_to_register:
+            log.warning(f"The permission {reg_permission} is not in the database. It will be *removed*")
+            tb_removed += [reg_permission]
+
+    permissions_to_register = list(set(permissions_to_register) - set(tb_removed))
+
```
ggsdc, line 72: There is something in the logic here that I do not seem to be able to understand.

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -55,6 +55,21 @@ def register_base_permissions_command(verbose):
         not in permissions_registered
     ]
 
+    for to_reg_permission in permissions_to_register:
+        if to_reg_permission in permissions_registered:
+            log.info(f"The permission to be registered {to_reg_permission} is already in the database")
+        if to_reg_permission not in permissions_registered:
+            log.info(f"The permission to be registered {to_reg_permission} is not registered. It will be registered")
+            permissions_registered.append(to_reg_permission)
```
ggsdc, line 63: Why append to the ones that already are on the database? Shouldn't you append to an empty list?

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -55,6 +55,21 @@ def register_base_permissions_command(verbose):
...
+
+    tb_removed = []
+    for i, reg_permission in enumerate(permissions_registered):
+        if reg_permission not in permissions_to_register:
+            log.warning(f"The permission {reg_permission} is not in the database. It will be *removed*")
+            tb_removed += [reg_permission]
+
+    permissions_to_register = list(set(permissions_to_register) - set(tb_removed))
```
ggsdc, line 71: I think this list is not the right one

cornflow-server/cornflow/commands/permissions.py
ggsdc, line 72: Where do you delete the permissions that have to be deleted?

PR #368, opened 2023-02-09
Added 'free' endpoint

marioncottard:


ggsdc (approved):

ggsdc:
I'm not 100% convinced that we should have more than the get methods, no POST or PUT and see what we can do about filters

marioncottard:
Okay, I'll delete them then. And I'll try and add the filters

cornflow-server/cornflow/shared/const.py
```diff
@@ -137,4 +137,6 @@
     (SERVICE_ROLE, POST_ACTION, "deployed-dag"),
     (SERVICE_ROLE, POST_ACTION, "instance"),
     (SERVICE_ROLE, POST_ACTION, "execution"),
+    (SERVICE_ROLE, POST_ACTION, "tables"),
+    (SERVICE_ROLE, POST_ACTION, "tables-detail"),
 ]
```
ggsdc, line 142: This permissions do not need to exist anymore

ggsdc (approved):

cornflow-server/cornflow/external_app/__init__.py
```diff
@@ -0,0 +1 @@
+from .model import *
```
ggsdc, line 1: Why this change?

ggsdc (changes requested):

cornflow-server/cornflow/external_app/__init__.py
marioncottard: It was so that the import_models()  method in cornflow-server/cornflow/shared/utils.py, that imports all of the models for the TablesEndpoint, could find the models of external_app. Otherwise they weren't getting imported

ggsdc: Ok, but that part from external_app is getting moved, so we will have to see how to adapt it for the external_app later on. I will make a fast commit to delete ok?

marioncottard: Okay

ggsdc (approved):

PR #372, opened 2023-02-10
Feature/add filter capacities get

AlejandroMus:


libs/core/cornflow_core/schemas/query.py
```diff
@@ -11,3 +11,14 @@ class BaseQueryFilters(Schema):
     offset = fields.Int(required=False, dump_default=0)
     creation_date_gte = fields.DateTime(required=False)
     creation_date_lte = fields.DateTime(required=False)
+    id = fields.Str(required=True)
+    name = fields.Str(required=True)
+    deletion_date_gte = fields.DateTime(required=False)
+    deletion_date_lte = fields.DateTime(required=False)
+    update_date = fields.DateTime(required=False)
```
ggsdc, line 18: The `update_date` should allow `gte` and `lte` as the `creation_date` and `deletion_date`

cornflow-server/cornflow/schemas/common.py
```diff
@@ -10,10 +10,19 @@ class QueryFilters(BaseQueryFilters):
...
     schema = fields.Str(required=False)
+    updated_at = fields.DateTime()
+    deleted_at = fields.DateTime()
+    user_name = fields.Str(required=True)
+    last_name = fields.Str(required=True)
+    email = fields.Str(required=True)
+    role_id = fields.Int(required=True)
+    url_rule = fields.Str(required=True)
+    description = fields.Str(required=True)
+
```
ggsdc, line 28: I think these changes are not needed. `BaseDataEndpointResponse` is used as the base of other schemas, so if you need to add fields, you should add them to the other schemas not these ones. For example you are making so that the instance, execution and case endpoints that have schemas that inherit from here should give back an `email` field as it is `required=True` so that does not make sense.

libs/core/cornflow_core/schemas/query.py
```diff
@@ -11,3 +11,14 @@ class BaseQueryFilters(Schema):
...
+    user_name = fields.Str(required=True)
+    last_name = fields.Str(required=True)
+    email = fields.Str(required=True)
+    role_id = fields.Int(required=True)
+    url_rule = fields.Str(required=True)
+    description = fields.Str(required=True)
```
ggsdc, line 24: Careful here, no query field should be `required=True` that can give a lot of problems. The four ones that we had are `required=False`

ggsdc (changes requested):

marioncottard:
I think you're also going to need to edit the get_all_objects() method of class BaseDataModel with the new filtering conditions

marioncottard:
Maybe also the get_all_objects() method from TraceAttributesModel, but I am not sure about that one: for now it doesn't do any filtering, I don't know if it is needed. I think it could be there just in case, though. @ggsdc ?

cornflow-server/cornflow/models/base_data_model.py
```diff
@@ -83,6 +105,26 @@ def get_all_objects(
...
+        if last_name:
+            query = query.filter(cls.last_name == last_name)
+        if email:
+            query = query.filter(cls.email == email)
+        if role_id:
+            query = query.filter(cls.role_id == role_id)
+        if url_rule:
+            query = query.filter(cls.url_rule == url_rule)
+        if description:
+            query = query.filter(cls.description == description)
```
ggsdc, line 127: I'm trying to think if there is a better way to do this....

ggsdc (dismissed):

ggsdc (approved):

PR #390, opened 2023-03-01
Bugfix/sprint7 t12

marioncottard:
Fixes integer validation issue when validating data with marshmallow: switch to validation with jsonschema


ggsdc (commented):
Do we use anywhere the `validate_and_continue` function now? Should we delete it?

marioncottard:
No, we don't use it anymore. I will delete it

ggsdc (approved):

PR #399, opened 2023-03-08
Added the indicators to the rostering DAG

ggsdc:
Added the following indicators for the rostering DAG:

* % of hours with just one employee.
* Mean demand that each employee has to cover
* Min demand that the employees have to cover.
* Max demand that the employees have to cover

### DIFF
--- a/cornflow-dags/DAG/rostering/core/experiment.py
+++ b/cornflow-dags/DAG/rostering/core/experiment.py
@@ -51,7 +51,7 @@ class Experiment(ExperimentCore):
     def get_indicators(self) -> dict:
         return {
             "objective_function": self.get_objective(),
-            "one_employee": self.get_one_employee_percentage(),
+            "only_one_employee_percentage": self.get_one_employee_percentage(),
             "mean_demand": self.get_mean_demand_employee(),
             "max_demand": self.get_max_demand_employee(),
             "min_demand": self.get_min_demand_employee(),
--- a/cornflow-dags/DAG/rostering/core/experiment.py
+++ b/cornflow-dags/DAG/rostering/core/experiment.py
@@ -49,7 +49,13 @@ class Experiment(ExperimentCore):
         return self.solution.get_working_hours()
 
     def get_indicators(self) -> dict:
-        return {"fo": self.get_objective()}
+        return {
+            "objective_function": self.get_objective(),
+            "one_employee": self.get_one_employee_percentage(),
+            "mean_demand": self.get_mean_demand_employee(),
+            "max_demand": self.get_max_demand_employee(),
+            "min_demand": self.get_min_demand_employee(),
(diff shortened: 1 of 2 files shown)


cornflow-dags/DAG/rostering/core/experiment.py
```diff
@@ -49,7 +49,13 @@ def get_objective(self) -> float:
         return self.solution.get_working_hours()
 
     def get_indicators(self) -> dict:
-        return {"fo": self.get_objective()}
+        return {
+            "objective_function": self.get_objective(),
+            "one_employee": self.get_one_employee_percentage(),
```
JaimeSotomayor, line 54: The name may not be very representative, I had to look it up. Maybe better "only_one_employee" or ""pct_only_one_employee", "%_only_one_employee". At least with the % or pct in the name

cornflow-dags/DAG/rostering/core/experiment.py
```diff
@@ -127,3 +133,40 @@ def check_skills_demand(self) -> list:
                 lambda v: {"timeslot": v[0], "id_skill": v[1], "number_missing": v[2]}
             )
         )
+
+    def get_one_employee_percentage(self) -> float:
+        """Returns the percentage of time slots where only one employee is working"""
+        ts_employee = self.solution.get_ts_employee()
+        return (
+            sum(1 for e in ts_employee.values() if len(e) == 1) / len(ts_employee) * 100
```
JaimeSotomayor, line 141: Although ts_employee coincides with the total number of time slots, I would divide by the total number of timeslots just in case there is no need to always have an employee

JaimeSotomayor (approved):
Minor comments, I approve because they are not even strictly necessary, just suggestions.

cornflow-dags/DAG/rostering/core/experiment.py
ggsdc: Changed

cornflow-dags/DAG/rostering/core/experiment.py
ggsdc: But then the timeslot should not be created, no? As the list of timeslots should be just the timeslots where there has to be _someone_ working

JaimeSotomayor: Of course, that's why the values are the same. But if the requirement were to change and instead of there always having to be someone only needed to cover the demand, the timeslot could be created, it could be necessary to cover the minimum number of hours

marioncottard (approved):

cornflow-dags/DAG/rostering/core/experiment.py
ggsdc: Ok, will change it, it should be an easy change

PR #405, opened 2023-03-09
Feature/more generic functions

hugolarzabal:
Possible new generic functions for cornflow core and cornflow client.

I do have several functions, endpoints and model which I tend to reuse between projects. So it may be better to have them directly in cornflow.

For the moment I did put all the functions in core in a "new_functions" directory.
The functions which are being considered usefull should later be moved to where it makes more sense.


ggsdc:
@marioncottard what do we do with this issue?

PR #414, opened 2023-03-15
Sprint 7

ggsdc:
All changes in sprint 7


cornflow-dags/tests/test_dags_af.py
```diff
@@ -30,39 +31,38 @@ def run_update_all_variables_until_finished(self):
...
             state = status.json()["state"]
             finished = state != "running"
             print("STATUS OF update_all_variables: {}".format(state))
         return client
 
+    def test_all_dags_exist(self):
+        names = [app.name for app in self.apps]
+        missing = set(EXISTING_APPS).difference(names)
+        print("Missing apps: {}".format(missing))
+        self.assertEqual(len(missing), 0)
```
marioncottard, line 44: I am not sure I see the point of this test. `EXISTING_APPS` and `names` are defined in equivalent ways, so the comparison really doesn't make sense to me

marioncottard (approved):

PR #436, opened 2023-03-31
Bugfix/tables endpoint imports

marioncottard:


ggsdc (approved):
Seems alright. Will test it out on my machine just to make sure

PR #437, opened 2023-03-31
Feature/mapping solver parameters

marioncottard:


ggsdc (approved):
Could you change any DAG to use these new methods in order to see that it works properly and that it can be set up as an example of how to do it

ggsdc (approved):

AlejandroMus (approved):
It is a bit strange to change timeLimit to time_limit when the rest of parameters follow the convention name1_name2. After talking with @marioncottard I know the reason, but still, it is not homogenous

hugolarzabal:
> It is a bit strange to change timeLimit to time_limit when the rest of parameters follow the convention name1_name2. After talking with @marioncottard I know the reason, but still, it is not homogenous

Yes, I do not understand it either. Why does it need to be changed?

marioncottard:
> > It is a bit strange to change timeLimit to time_limit when the rest of parameters follow the convention name1_name2. After talking with @marioncottard I know the reason, but still, it is not homogenous
> 
> Yes, I do not understand it either. Why does it need to be changed?

It is because cornflow-app allows you to select a time limit when creating an execution and when you do, it is passed to the config as "timeLimit". If we change the name to time_limit here, we would need to change it in the interface too if we want this method to be compatible with the use of the interface. It is doable, of course, but I figured it was easier to just leave it as timeLimit here

hugolarzabal:
> > > It is a bit strange to change timeLimit to time_limit when the rest of parameters follow the convention name1_name2. After talking with @marioncottard I know the reason, but still, it is not homogenous
> > 
> > 
> > Yes, I do not understand it either. Why does it need to be changed?
> 
> It is because cornflow-app allows you to select a time limit when creating an execution and when you do, it is passed to the config as "timeLimit". If we change the name to time_limit here, we would need to change it in the interface too if we want this method to be compatible with the use of the interface. It is doable, of course, but I figured it was easier to just leave it as timeLimit here

hum, I think it would be more consistent to change it in cornflow app then. Having this inconsistency in one project would not be a big problem but I do not like the idea to have it in every future project.

libs/client/cornflow_client/core/experiment.py
```diff
@@ -81,3 +83,59 @@ def schema_checks(self) -> dict:
             the method ExperimentCore.check_solution()
         """
         raise NotImplementedError()
+
+    @staticmethod
+    def get_solver_config(config, lib="pyomo", default_solver="cbc", remove_unknown=False):
+        """
```
hugolarzabal, line 89: remove_unknown is not in the docstring

libs/client/cornflow_client/core/experiment.py
```diff
@@ -81,3 +83,59 @@ def schema_checks(self) -> dict:
...
+            conf = {
+                mapping[k, lib, solver]: v for k, v in config.items() if (k, lib, solver) in mapping
+            }
+        else:
+            conf = {
+                mapping.get((k, lib, solver), k): v for k, v in config.items()
+            }
+            conf.pop("solver", None)
+            conf.pop("solver_config", None)
+        if config.get("solver_config"):
```
hugolarzabal, line 135: I do not get this part. Why do you copy solver config? Why do you go through the function a second time instead of simply updating conf?

libs/client/cornflow_client/core/experiment.py
```diff
@@ -81,3 +83,59 @@ def schema_checks(self) -> dict:
...
+
+        solver = SOLVER_CONVERTER.get(solver)
+        if solver is None:
+            print(
+                "The solver doesn't correspond to any solver in the parameters mapping. "
+                f"Default solver {default_solver} will be used."
+            )
+            solver = default_solver
+
+        if remove_unknown:
```
hugolarzabal, line 125: I do not really understand the use of remove_unknown

hugolarzabal (changes requested):

libs/client/cornflow_client/core/experiment.py
marioncottard: You're right. It is now

libs/client/cornflow_client/core/experiment.py
marioncottard: If it is True, the method removes the parameters that are not in PARAMETER_SOLVER_TRANSLATING_MAPPING from the config. Otherwise it just doesn't translate it

libs/client/cornflow_client/core/experiment.py
```diff
@@ -81,3 +83,59 @@ def schema_checks(self) -> dict:
...
+            solver = solver.split(".")[1]
+
+        solver = SOLVER_CONVERTER.get(solver)
+        if solver is None:
+            print(
+                "The solver doesn't correspond to any solver in the parameters mapping. "
+                f"Default solver {default_solver} will be used."
+            )
+            solver = default_solver
+
```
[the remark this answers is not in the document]
marioncottard: That makes sense, thanks

libs/client/cornflow_client/core/experiment.py
marioncottard: Makes sense too

libs/client/cornflow_client/constants.py
```diff
@@ -116,21 +116,25 @@
...
     ('is_mip', 'pulp', 'cbc') : 'mip' ,
     ('abs_gap', 'pulp', 'cbc') : 'gapAbs' ,
     ('rel_gap', 'pulp', 'cbc') : 'gapRel' ,
     ('cutoff', 'pulp', 'cbc') : 'cuts' ,
-    ('time_limit', 'pulp', 'cbc') : 'timeLimit' ,
+    ('time_limit', 'pulp', 'cbc'): 'timeLimit',
+    ('timeLimit', 'pulp', 'cbc') : 'timeLimit' ,
     ('threads', 'pulp', 'cbc') : 'threads' ,
     ('presolve', 'pulp', 'cbc') : 'presolve' ,
+    ('msg', 'pulp', 'cbc'): 'msg',
```
ggsdc, line 131: Some of the commas here have a space that I do not think is needed, but I do not want to drag this PR too much.

ggsdc (approved):

hugolarzabal (approved):

PR #438, opened 2023-03-31
Feature/error handler

marioncottard:


cornflow-server/cornflow/tests/ldap/test_ldap_authentication.py
```diff
@@ -116,4 +116,7 @@ def test_deactivated_sign_up(self):
             follow_redirects=True,
             headers={"Content-Type": "application/json"},
         )
+        print(response)
+        print(response.__dict__)
+        print(response.json)
```
ggsdc, line 121: This should be left out

libs/core/cornflow_core/exceptions/exceptions.py
```diff
@@ -4,7 +4,9 @@
 """
 from flask import jsonify
 from webargs.flaskparser import parser
+from werkzeug.exceptions import InternalServerError
```
ggsdc, line 7: `InternalServerError` is not used anywhere?

ggsdc (commented):
Is this PR finished?

cornflow-server/cornflow/tests/ldap/test_ldap_authentication.py
marioncottard: Done

libs/core/cornflow_core/exceptions/exceptions.py
marioncottard: You're right

PR #442, opened 2023-04-05
Feature/alarms

marioncottard:
Adapted from https://github.com/baobabsoluciones/cornflow/pull/405.


cornflow-server/cornflow/config.py
```diff
@@ -92,6 +95,7 @@ class Testing(DefaultConfig):
     AIRFLOW_PWD = os.getenv("AIRFLOW_PWD", "admin")
     OPEN_DEPLOYMENT = 1
     LOG_LEVEL = int(os.getenv("LOG_LEVEL", 10))
+    ALARMS_ENDPOINTS = 1
```
ggsdc, line 98: Should be 0 as default?

cornflow-server/cornflow/tests/unit/test_main_alarms.py
```diff
@@ -0,0 +1,69 @@
+"""
+
+"""
+# TODO: Remove this
+import sys, os
+prev_dir = os.path.join(os.path.dirname(__file__), "..", "..", "..")
+print(prev_dir)
```
ggsdc, line 7: This print has to go out

cornflow-server/cornflow/tests/unit/test_main_alarms.py
```diff
@@ -0,0 +1,69 @@
...
+sys.path.insert(1, prev_dir)
```
ggsdc, line 8: Why do you need this?

docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
       - CORNFLOW_DB_USER=cornflow
       - CORNFLOW_DB_PASSWORD=cornflow
       - CORNFLOW_DB=cornflow
+      - CF_ALARMS_ENDPOINT=1
```
ggsdc, line 21: This should be zero

cornflow-server/cornflow/tests/unit/test_main_alarms.py
marioncottard: Because this way I can run the tests directly with Pycharm

ggsdc: That is weird, you should be able to run the tests without the need for this.

ggsdc (approved):
Once is out of draft phase and in sync with development we can merge

PR #453, opened 2023-04-12
Release/v1.0.4

ggsdc:
This should have all the changes introduced during sprint 8.
Please everyone review that everything is up to order.
I will _soon_ update the version codes for deployment.

- [x] Update cornflow version
- [x] Update cornflow client version
- [x] Update cornflow core version
- [x] Update cornflow changelog
- [x] Update cornflow client changelog
- [ ] Update private DAGS


ggsdc:
With these PR most of you open PR (@hugolarzabal ) should be covered. Is that right @marioncottard ?

hugolarzabal (commented):
a few comments:
- what is the tables endpoint?
- I that that some tools that I suggested are not in that PR, is it planned to include some of them later?
- is it planned to include later a generic endpoint for creating, posting and solving an instance?

hugolarzabal (approved):

AlejandroMus (dismissed):

ggsdc:
> a few comments:
> 
>     * what is the tables endpoint?
> 
>     * I that that some tools that I suggested are not in that PR, is it planned to include some of them later?
> 
>     * is it planned to include later a generic endpoint for creating, posting and solving an instance?

The tables endpoint allows to query in a generic form any table in the database.
The rest of the points it should be @marion who answers

marioncottard:
Hi, 

> * I that that some tools that I suggested are not in that PR, is it planned to include some of them later?

I reviewed the methods that you suggested and separated them between three categories:
- the ones that could and would be reused
- the ones that were too specific to be reused (for instance, the methods to create alarms. I think the way alarms are used will really depend on the project)
- the ones that were actually pretty simple methods, in a way that I figured it was almost easier to implement them directly than to have a method stored in cornflow (for example, the as_list and between methods). 

I only kept the ones that fell into the first category, because in my opinion, it is best to not crowd cornflow with a lot of methods that are not generic enough to be reused, even though they are useful in some specific projects

> * is it planned to include later a generic endpoint for creating, posting and solving an instance?

Not really. From what I get. the projects that use these kinds of endpoints generally have a data flow that is very specific to those projects, so a generic endpoint would basically be empty, or have methods that would be systematically overwritten

ggsdc:
@AlejandroMus @hugolarzabal @marioncottard I need an approval....

marioncottard (approved):

PR #480, opened 2023-10-05
Apply offset only if not null

marioncottard:


ggsdc:
Why did you close it?

marioncottard:
Because I can't figure out why this breaks the tests and I can do it another way

ggsdc:
The failing test are not your fault, I have been fighting with them the
past days. If you run them again it should work eventually

marioncottard:
Okay, I'll try then

ggsdc:
Could you solve the conflicts (due to me commenting a big part of the code to check if the test now pass)

marioncottard:
I've left commented the tests that you commented, because even though they passed on local, they wouldn't pass here

PR #485, opened 2023-10-13
Now responses go back in the intended order instead of in alphabetical order

ggsdc:
Json response are not sorted alphabetically and should go with the order in the original dict


ggsdc:
@CodiumAI-agent /review

marioncottard (approved):

PR #488, opened 2023-10-20
fix add and config ssh key into user home dir

ampuerin:
The problem is that the user directory of the solvers image is not the same as where the ssh key was added and configured to install dependencies from private repositories.

Now we have to use as reference in airflow Dockerfile the `${HOME}/.ssh` directory and not `$AIRFLOW_HOME/.ssh` directory


ampuerin:
This PR can be closed without merge because solution is not valid.

PR #525, opened 2024-04-22
Feature/examples

### LINKED ISSUE #513: Improve the example endpoint and maybe join it with the schema one

I don't like the current way the `example` endpoint is implemented for various reasons:

1. You cannot download just one example, you download all of them each time.
2. The instances and their solution are not well organized, they match through the name which is very precarious.
3. The endpoint (example) seems very specific. We could generalize something to be a App-specific endpoint where we have the option of getting (a) the schema, (b) examples, (c) something else, for example, a description of the app, a logo, etc.

1. There should be two endpoints: one for getting a list of examples for a given App and the other one should download the details of only one example.
2. Instead of downloading Instances and Solutions, one should be able to download Experiments that have a Solution=null for the instances that are not solved.
3. A unified API for dags/Apps.

**Describe alternatives you've considered**

downloading all examples every time I want one example.

ggsdc:
Start the changes to implement a better example data handling on cornflow.

For now the changes include:

- new structure of the test cases property.
- refactor of the cornflow-dags examples and unit tests.
- adaptation on the update_all_schemas default DAG.

Closes #513


pchtsp:
this looks good, when can we expect to have it in the merged/deployed in the development server?

ggsdc:
If everything goes according to plan I expect for it to be on the
development server for some time next week along some other changes



Guillermo González-Santander de la Cruz

JEFE DE PROYECTOS

T  | +34 915 337 301

E  | ***@***.***

W | baobabsoluciones.es

pchtsp:
Hello, I'm checking again these changes. It seems the endpoints have not changed. Meaning: we cannot get the list of example names and then, the selected example through the REST API.

PR #526, opened 2024-05-09
feature: added log_json to ExecutionEndpoint and ExecutionDataEndpoint responses

claudio-rodriguez-dev:


JaimeSotomayor (approved):

cornflow-server/cornflow/schemas/execution.py
```diff
@@ -129,6 +133,7 @@ class ExecutionStatusEndpointUpdate(Schema):
 class ExecutionDataEndpointResponse(ExecutionDetailsEndpointResponse):
     data = fields.Raw()
     checks = fields.Raw()
+    basic_log_json = fields.Nested(BasicLogSchema, attribute="log_json")
```
ggsdc, line 136: Why do we modify this schema as well?

cornflow-server/cornflow/schemas/execution.py
```diff
@@ -114,6 +114,10 @@ def get_indicators(self, obj):
     indicators = fields.Method("get_indicators")
 
 
+class ExecutionDetailsWithIndicatorsAndLogResponse(ExecutionDetailsEndpointWithIndicatorsResponse):
+    basic_log_json = fields.Nested(BasicLogSchema, attribute="log_json")
```
ggsdc, line 118: I would not call this field `basic_log_json`. I think with a name as `log` it should be enough, in this case it only bring some information instead of the full log, but it should be enough.

ggsdc, line 118: Why not explode this and give back the two or three fields that we want as new fields instead of a dict inside a dict?

This can be done with a simple `fields.Method(callable)` where callable is the function that extracts the desired value from the log that can be defaulted to some other value in case it is needed.

cornflow-server/cornflow/schemas/solution_log.py
```diff
@@ -68,3 +68,9 @@ class LogSchema(Schema):
     nodes = fields.Int(**log_options)
     progress = fields.Nested(ProgressSchema, required=False)
     cut_info = fields.Raw(**log_options)
+
+
+class BasicLogSchema(Schema):
+    status = fields.Str(**log_options)
+    status_code = fields.Int(**log_options)
+    sol_code = fields.Int(**log_options)
```
ggsdc, line 76: Why not make the other `LogSchema` inherit from this new schema?

cornflow-server/cornflow/tests/custom_test_case.py
```diff
@@ -199,7 +203,7 @@ def get_keys_to_check(self, payload):
         return payload.keys()
 
     def get_one_row(
-        self, url, payload, expected_status=200, check_payload=True, token=None
+            self, url, payload, expected_status=200, check_payload=True, token=None, keys_to_check: List[str] = None
```
ggsdc, line 206: Are you running black? This formatting change seems odd

cornflow-server/cornflow/tests/custom_test_case.py
```diff
@@ -227,13 +233,13 @@ def get_no_rows(self, url, token=None):
...
-        expected_status=200,
-        check_payload=True,
-        token=None,
+            self,
+            url,
+            change,
+            payload_to_check,
+            expected_status=200,
+            check_payload=True,
+            token=None,
```
ggsdc, line 242: Same to this formatting change. If I run black locally it changes back this changes, which version of black are you running?

cornflow-server/cornflow/tests/custom_test_case.py
```diff
@@ -264,7 +270,7 @@ def update_row(
         return row.json
 
     def patch_row(
-        self, url, json_patch, payload_to_check, expected_status=200, check_payload=True
+            self, url, json_patch, payload_to_check, expected_status=200, check_payload=True
```
ggsdc, line 273: Same here

cornflow-server/cornflow/tests/custom_test_case.py
```diff
@@ -322,7 +328,7 @@ def str_method(self, idx, string: str):
         self.assertEqual(str(row), string)
 
     def cascade_delete(
-        self, url, model, payload, url_2, model_2, payload_2, parent_key
+            self, url, model, payload, url_2, model_2, payload_2, parent_key
```
ggsdc, line 331: Same here

cornflow-server/cornflow/tests/custom_test_case.py
```diff
@@ -410,9 +416,9 @@ def url_with_query_arguments(self):
                 return self.url
             else:
                 return (
-                    self.url
-                    + "?"
-                    + "&".join(["%s=%s" % _ for _ in self.query_arguments.items()])
+                        self.url
+                        + "?"
+                        + "&".join(["%s=%s" % _ for _ in self.query_arguments.items()])
```
ggsdc, line 421: Same here

cornflow-server/cornflow/tests/unit/test_executions.py
```diff
@@ -115,7 +115,9 @@ def test_new_execution_no_instance(self):
         self.assertTrue("error" in response.json)
 
     def test_get_executions(self):
-        self.get_rows(self.url, self.payloads)
+        keys_to_check = ['instance_id', 'schema', 'indicators', 'id', 'config', 'data_hash', 'basic_log_json',
+                         'created_at', 'state', 'user_id', 'message', 'name', 'description']
+        self.get_rows(self.url, self.payloads, keys_to_check=keys_to_check)
```
ggsdc, line 120: Should it make sense to at least add this keys to check argument to all the test where the payload is not checked?

ggsdc (changes requested):

cornflow-server/cornflow/schemas/execution.py
claudio-rodriguez-dev: I did it the same as the ExecutionLogEndpointResponse schema, if log_json is a json it seems cleaner to me to parse it with a schema than to make a method that extracts three fields. In case the log schema was changed at some point, that would be enough and there would be no need to go to the method to change the fields.

ggsdc (approved):

PR #510, opened 2024-01-31
Added missing migration

ggsdc:

PR #506, opened 2024-01-11
Move docs folder

ggsdc:
Solves issue #495 

Moved documentation from inside cornflow-server folder to root folder.
Updated documentation.
Changed documentation theme.

PR #491, opened 2023-10-31
Translate configuration dictionaries according to solver in all dags

marioncottard:

PR #533, opened 2024-05-15
Release/v1.1.0

ggsdc:
New version of cornflow-server and cornflow-client with the following changes:

- Custom token duration. It defaults to 24 hours. PR #507
- Fixed error that can appear on log in. PR #509
- Added password rotation capabilities. Ir defaults to 120 days. Currently it only notifies if the password should be changed, it does not force it just yet. PR #508. Migration on PR #510 
- Better translation for different modelling libraries and solvers. PR #491
- Several libraries updates.
- Updated license endpoint. PR #519 
- Moved docs folder #506.
- Fixed bug that accepted empty passwords as valid #512 
- Changed tests.
-


SECURITY.md
```diff
@@ -0,0 +1,19 @@
...
+
+## What should be and should NOT be reported?
+
+**Only** use the security e-mail address to report undisclosed security vulnerabilities in cornflow and to manage the process of fixing such vulnerabilities. We do not accept regular bug reports or other security-related queries at this address. We will ignore mail sent to this address that does not relate to an undisclosed security problem in the cornflow project. Please follow regular communication channels for inquiries, questions and other discussions related to the process or issues.
+
+Specifically, we will ignore results of security scans that contain a list of dependencies of cornflow with dependencies in cornflow docker reference image.
+
+## How to report the issue?
+
+Please send one plain-text email for each vulnerability you are reporting including an explanation of how it affects Airflow security. We may ask that you resubmit your report if you send it as an image, movie, HTML, or PDF attachment when you could as easily describe it with plain text.
```
pchtsp, line 19: it says "Airflow" here, I guess we forgot to replace it to "Cornflow"? If we're using their text, maybe we should say we're basing it on theirs? I'm not sure.

ggsdc: I have missed yes. I think there is no problem reusing part of their security policy right?

marioncottard (approved):

libs/client/cornflow_client/core/instance_solution.py
```diff
@@ -335,22 +333,42 @@ def add_time_to_ts(ts: datetime, weeks=0, days=0, minutes=0, seconds=0) -> datet
         return ts + timedelta(days=7 * weeks + days, minutes=minutes, seconds=seconds)
 
     @staticmethod
-    def add_time_to_date_string(string: str, weeks=0, days=0, minutes=0, seconds=0) -> str:
+    def add_time_to_date_string(
```
pchtsp, line 336: why do we have all these dates functions in the instance_solution class? shouldn't we have them all under a time package that can be itself used from the instance? or a different package/ class that can be called from outside instance? With the current way of working the instance is polluted with a huge number of date-specific methods that are not necessarily relevant for all problems.

JaimeSotomayor (approved):

libs/client/cornflow_client/core/instance_solution.py
ggsdc: We have these functions in another library which we could use as a dependency but I haven't got around in doing the change.

PR #537, opened 2024-05-21
Feature/examples

ggsdc:


pchtsp:
What if there are two examples with the same name? We just return the first one I guess? Are we imposing a unique name? Since we create them I guess there's not much risk.

ggsdc:
> What if there are two examples with the same name? We just return the first one I guess? Are we imposing a unique name? Since we create them I guess there's not much risk.

I would say that yes, It is up to the developer to do not give the same name to two examples, in that case only the first one will be given back.

PR #562, opened 2024-11-04
Small bugfix for long solving sessions

ggsdc:
@marioncottard you ahve encountered a bug that could be solved with this right?

### DIFF
--- a/libs/client/cornflow_client/airflow/dag_utilities.py
+++ b/libs/client/cornflow_client/airflow/dag_utilities.py
@@ -226,6 +226,8 @@ def cf_solve(fun, dag_name, secrets, **kwargs):
     except NoSolverException as e:
         if config.get("msg", True):
             print("No solver found !")
+        # We reconnect in case the solver has been more than 24 hours solving before the error is raised
+        client = connect_to_cornflow(secrets)
         try_to_save_error(client, exec_id, -1)
         client.update_status(exec_id, {"status": -1})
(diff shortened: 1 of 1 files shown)


libs/client/cornflow_client/airflow/dag_utilities.py
```diff
@@ -226,13 +226,17 @@ def cf_solve(fun, dag_name, secrets, **kwargs):
     except NoSolverException as e:
         if config.get("msg", True):
             print("No solver found !")
+        # We reconnect in case the solver has been more than 24 hours solving before the error is raised
+        client = connect_to_cornflow(secrets)
```
AlejandraGalan, line 230: Would it be better to move the connect call inside try_to_save_error and pass secrets as an argument instead of client? This way, we could avoid creating a new connect call each time it's used.

AlejandraGalan (commented):
Small suggestion

libs/client/cornflow_client/airflow/dag_utilities.py
ggsdc: No, because the method `try_to_save_airflow_log()` also needs the client updated to be able to save the log as well. So if we update it inside the `try_to_save_error()` it would not be updated for the other method.

marioncottard:
We had a problem in one of the projects, yes, but I was thinking of increasing the TOKEN_DURATION variable to solve it. Because with this change, the error and log would be saved, but not the solution.

marioncottard (approved):

Issue #565, opened 2024-11-04
Add automatic check detection for Instance and Experiment class


ggsdc:
Default `check` method on `Instance` and `Experiment` should be able to detect automatically the checks that are created based on the method naming and run them all to return the desired checks.

This has done on a previous project and has to be migrated. For more information talk to @hugolarzabal or @bwronska

hugolarzabal:
My idea on that one would be to have the following:

- have a CheckInstanceSolution class with some general methods.
- on a project, create a CheckInstance and CheckSolution subclass.
- these classes can store the instance and solution  as attributes in order to access their data and methods.

Some ideas for the generic methods of those classes can be:

- have a method to launch all checks (methods with name starting with "check_".
- have a method to generate errors or warning depending on the results of the checks.
- have a method to generate alarms to be written in cornflow depending on the results of the checks.

I will implement those ideas in one of our current projects and then create a PR to cornflow.
Any idea or suggestion is welcome.

PR #571, opened 2024-11-18
add a class to check the instance or solution.

hugolarzabal:
Add a new class for creating checks on instance and solution and generating the corresponding alarms.

closes https://github.com/baobabsoluciones/cornflow/issues/565


ggsdc (dismissed):
There is a main think that I do not lik about these changes and that I think should be changed. this would be the *only* part inside the core parts of cornflow-client (instance, solution, experiment, etc..) that has a connection to cornflow. 

Why?

I do not think this is needed. I am going to change it.

bwronska (approved):

ggsdc:
Ok, so now it should be a smaller version of what we currently have on other projects.
By default both the instance and experimetn class will run all methods that start by the keyword `check_`.
It will accept any check that either returns a list or a dict of warnings / errors (that should be defined on the jsonschema).
All the methods: get_check_methods, launch_all_checks or check can be overrriden if more custom support is needed.

@hugolarzabal I have saved you work ona  separate branch just in case.
@bwronska please review comparing with your current implementations.

bwronska (dismissed):
I think this is a simpler implementation than the one I currently have

bwronska:
> > I think this is a simpler implementation than the one I currently have
> 
> Yes, but that is the idea, as not everyone uses the Table class for example, or divides the checks into errors and warnings on the code level, so I think this should cover the basic functionalities.

Yes, I meant that for me, it is an advantage

hugolarzabal:
I am not sure if I like the fact that those methods are in instance and experiment core.
In the one hand, it is simpler not to create a new class but on the other hand, that means that all the checks will be in instance and experiment together with many other methods to prepare the instance and solution.

hugolarzabal:
Also, I think it could be good to add some method to handle warning and error since this way of differentiating checks problems is already in the check schema and in the applicationCore with is_warning property.

In one project, I do have something like this:
```
    def _get_check_criticality(self, check_name):
        """
        Get criticality level from schema for a given check
        
        :param check_name: Name of the check method
        :return: 2 if is_warning is True, 1 otherwise
        """
        check_schema = self.inst.schema_checks["properties"].get(check_name, {})
        return 2 if check_schema.get("is_warning", False) else 1`
```

ggsdc:
> ment together with many other methods to prep

This is not actually needed. If you want to separate the checks is easily done:

``` python

class InstanceChecks(Instance):

  In this class you implement all the check_ methods


class Instance(InstanceCore):

  def check(self):
    instance_checks = InstanceChecks.from_instance(self)
    return instance_checks.launch_all_checks()

```

ggsdc:
> Also, I think it could be good to add some method to handle warning and error since this way of differentiating checks problems is already in the check schema and in the applicationCore with is_warning property.
> 
> In one project, I do have something like this:
> 
> ```
>     def _get_check_criticality(self, check_name):
>         """
>         Get criticality level from schema for a given check
>         
>         :param check_name: Name of the check method
>         :return: 2 if is_warning is True, 1 otherwise
>         """
>         check_schema = self.inst.schema_checks["properties"].get(check_name, {})
>         return 2 if check_schema.get("is_warning", False) else 1`
> ```

Why would you want this criticality level? You already know, thanks to the schema if they are errors, so the solver should not be run, or if they are warning, in which case they come back and the solver does run

hugolarzabal:
> > I am not sure if I like the fact that those methods are in instance and experiment core.
> > In the one hand, it is simpler not to create a new class but on the other hand, that means that all the checks will be in instance and experiment together with many other methods to prepare the instance and solution.
> 
> This is not actually needed. If you want to separate the checks is easily done:
> 
> ```python
> class InstanceChecks(Instance):
> 
>   In this class you implement all the check_ methods
> 
> 
> class Instance(InstanceCore):
> 
>   def check(self):
>     instance_checks = InstanceChecks.from_instance(self)
>     return instance_checks.launch_all_checks()
> ```

That 's a circular import (I know you can avoid the error by having some import statement in the method but this is quite ugly)

hugolarzabal:
> > Also, I think it could be good to add some method to handle warning and error since this way of differentiating checks problems is already in the check schema and in the applicationCore with is_warning property.
> > In one project, I do have something like this:
> > ```
> >     def _get_check_criticality(self, check_name):
> >         """
> >         Get criticality level from schema for a given check
> >         
> >         :param check_name: Name of the check method
> >         :return: 2 if is_warning is True, 1 otherwise
> >         """
> >         check_schema = self.inst.schema_checks["properties"].get(check_name, {})
> >         return 2 if check_schema.get("is_warning", False) else 1`
> > ```
> 
> Why would you want this criticality level? You already know, thanks to the schema if they are errors, so the solver should not be run, or if they are warning, in which case they come back and the solver does run

For logging a warning or an error or for creating an alarm in the UI.

As it is now, the app is completely silent when stopping the resolution because of the checks, we need to be able to log something about the problem.

ggsdc:
> > > Also, I think it could be good to add some method to handle warning and error since this way of differentiating checks problems is already in the check schema and in the applicationCore with is_warning property.
> > > In one project, I do have something like this:
> > > ```
> > >     def _get_check_criticality(self, check_name):
> > >         """
> > >         Get criticality level from schema for a given check
> > >         
> > >         :param check_name: Name of the check method
> > >         :return: 2 if is_warning is True, 1 otherwise
> > >         """
> > >         check_schema = self.inst.schema_checks["properties"].get(check_name, {})
> > >         return 2 if check_schema.get("is_warning", False) else 1`
> > > ```
> > 
> > 
> > Why would you want this criticality level? You already know, thanks to the schema if they are errors, so the solver should not be run, or if they are warning, in which case they come back and the solver does run
> 
> For logging a warning or an error or for creating an alarm in the UI.
> 
> As it is now, the app is completely silent when stopping the resolution because of the checks, we need to be able to log something about the problem.

The instance checks and solution checks get stored on the database with the instance or the execution, for the instance the checks can even be run before the execution to be able to check if the data is coherent before even trying to run the solver and if errors or warning are found are showed on the UI.

For the solution is the same.

This is all on the cornflow-ui interface, where would you want it to log it? BEcause either you have it on the database available through the REST API or you have it on the cornflow-ui interface.,

libs/client/cornflow_client/core/experiment.py
```diff
@@ -93,24 +93,60 @@ def data_checks(self) -> dict:
...
         :return: a dictionary of dictionaries. Each dictionary represents one type of error. Each of the elements
           inside represents one error of that particular type.
         """
-        pass
+        return self.launch_all_checks()
 
-    def check_solution(self) -> dict:
+    def check_solution(self) -> Dict[str, Union[List, Dict]]:
         """
         Mandatory method
```
AlejandraGalan, line 107: Now that this method has an implementation, is it still necessary to consider it mandatory? Or should its functionality be aligned with the 'check' method?

libs/client/cornflow_client/core/experiment.py
```diff
@@ -93,24 +93,60 @@ def data_checks(self) -> dict:
             )
         return checks
 
-    # TODO: make this method abstract on cornflow-client 2.0.0
-    def check(self) -> Dict[str, Dict]:
+    def check(self) -> Dict[str, Union[List, Dict]]:
         """
-        Mandatory method
+        Method that runs all the checks on the class.
```
AlejandraGalan, line 98: Should we explicitly document that the 'check' method can be overridden for customization,
similar to how we do it in the 'check' method in InstanceCore class?

AlejandraGalan (dismissed):

ggsdc:
@AlejandraGalan @bwronska @hugolarzabal @marioncottard any last comments?

libs/client/cornflow_client/core/application.py
```diff
@@ -188,7 +188,7 @@ def solve(
             .keys()
         )
         instance_errors = instance_checks.kfilter(lambda k: k not in warnings_tables)
-        if any([len(error_table) for error_table in instance_errors.values()]):
+        if len(instance_errors) > 0:
             log = dict(
```
marioncottard, line 192: I think the old formulation, `any([len(error_table) for error_table in instance_errors.values()])`, was better, because it didn't generate an error when the check() method returned empty lists for some checks (for example: `{"check_error_foreign_keys: []}`).  I've seen that in InstanceCore.launch_all_checks, you only return the check results if the list is not empty, but some other projects might still override the check() method and return empty lists, and I think CornFlow should be able to handle this possibility

ggsdc: You are right, i will change it back.

marioncottard (approved):

Issue #567, opened 2024-11-05
Add the ability to downgrade migrations from the cornflow cli


ggsdc:
**Is your feature request related to a problem? Please describe.**
If a downgrade need to be done (mainly on deployments with external apps) there is no option to do it through the cornflow cli

**Describe the solution you'd like**
Add the option as we have with the uppgrade and migrate

**Describe alternatives you've considered**
None

Issue #496, opened 2024-01-10
Some commands do not work properly when using them inside a docker container


ggsdc:
The testing of the commands has to be improved and the commands should be able to work inside the cornflow container in a standalone distribution and in one with an external app.

PR #574, opened 2024-12-02
Bugfix/review commands

marioncottard:
Fixes https://github.com/baobabsoluciones/cornflow/issues/496. Note that for the commands to work properly for external apps, the environment variable `DEFAULT_POSTGRES` needs to be defined.
Closes https://github.com/baobabsoluciones/cornflow/issues/567.


ampuerin (approved):

cornflow-server/cornflow/cli/migrations.py
```diff
@@ -41,6 +41,20 @@ def upgrade_migrations():
...
+    app = get_app()
+    external = int(os.getenv("EXTERNAL_APP", 0))
+    if external == 0:
+        path = "./cornflow/migrations"
+    else:
+        path = f"./{os.getenv('EXTERNAL_APP_MODULE', 'external_app')}/migrations"
+
+    with app.app_context():
+        migration_client = Migrate(app=app, db=db, directory=path)
+        downgrade()
```
ggsdc, line 55: For the downgrade you should pass a migration file name or version to downgrade to that version. 
We could also do that for the upgrade as well.

ggsdc (changes requested):
Small change, but I guess that you have tested that the cornflow cli is fully usuable from inside the container right? The main use for this will be the user creation / handling and we can addmore commands in that sense

ggsdc:
@marioncottard when do you think are you going to be able to tackle these changes?

PR #577, opened 2024-12-04
Bugfix/jsonschema

### LINKED ISSUE #575: Solution data is not validated against json schema after the solver finishes
**Describe the bug**
The current implementation of how the DAGs are deployed in airflow makes it so that the solutiond ata does not pass through validation in the json schema.

This should happen in the boilerplate for `ApplicationCore` `solve` method before giving back the control to the `cf_solve` method in `cornflow_client.airflow.dag_utilities`.

If we implement it there we should make sure of two things:

- [ ] 1. That the json schema validation fo the solution is run on cornflow-dags tests and possible test in our custom internal repositories.
- [ ] 2. That this should not ahve an impact of currently deployed DAGs on production environments.

**To Reproduce**
Solve any DAG on airflow. If solution does not comply with schema isntead of a `BadSolution` exception we get a different that 200 status code for writing the solution on cornflow.

**Expected behavior**
Either raise the `BadSolution` or do not raise anything at all.

**Metadata (please complete the following information):**
- cornflow-client==1.0.16 (but this should happen on newer versions as well.

**Additional context**
@JaimeSotomayor can give more context if needed.

### LINKED ISSUE #576: Solution and instance checks are not validated against the json schema
**Describe the bug**
The current implementation of how the DAGs are deployed in airflow makes it so that the instance and solution checks does not pass through validation in the json schema.

This should happen in the boilerplate for `ApplicationCore` `solve` method before giving back the control to the `cf_solve` method in `cornflow_client.airflow.dag_utilities`.

**To Reproduce**
If any modification si introduced on the check calculation the error is found on cornflow, not on airflow.

**Expected behavior**
To fail on the airflow side if data does not pass the json schema validation.

**Metadata (please complete the following information):**
 - cornflow-client==1.0.16 (but should happen on newer versions)

ggsdc:
This PR should implement json schmea validation for instance, solution, instance checks and solution checks on the airflow side.

It should fix #575 and fix #576

### DIFF
--- a/libs/client/cornflow_client/core/application.py
+++ b/libs/client/cornflow_client/core/application.py
@@ -1,6 +1,7 @@
 """
 
 """
+
 # Partial imports
 from abc import ABC, abstractmethod
 from timeit import default_timer as timer
@@ -19,6 +20,8 @@ from cornflow_client.constants import (
     BadConfiguration,
     BadSolution,
     BadInstance,
+    BadInstanceChecks,
+    BadSolutionChecks,
 )
 from .experiment import ExperimentCore
 
@@ -181,6 +184,12 @@ class ApplicationCore(ABC):
 
         instance_checks = SuperDict(inst.check())
 
+        inst_errors = inst.validate_checks(instance_checks)
+        if inst_errors:
+            raise BadInstanceChecks(
+                f"The instance does not match the schema:\n{inst_errors}"
+            )
+
         warnings_tables = (
             SuperDict.from_dict(inst.schema_checks)["properties"]
             .vfilter(lambda v: v.get("is_warning", False))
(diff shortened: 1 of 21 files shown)


libs/client/cornflow_client/core/application.py
```diff
@@ -181,6 +184,12 @@ def solve(
 
         instance_checks = SuperDict(inst.check())
 
+        inst_errors = inst.validate_checks(instance_checks)
```
JaimeSotomayor, line 187: i wouldn't use "inst_errors" again because it can be confusing with the errors of the instance schema. I think "inst_checks_errors" is better. 
And the same with "inst_errors". I would prefer "inst_squema_errors"

If anything, two different names. If you like to keep "inst_errors" for me that would be for the checks. And for the squema of the instance i would use inst_squema_errors

ggsdc: I have changed the way it is done. See if you like it.

ampuerin (dismissed):

libs/client/cornflow_client/core/instance.py
```diff
@@ -12,8 +16,25 @@ class InstanceCore(InstanceSolutionCore, ABC):
...
+        Method that executes the InstanceCore.check() method and validates the result against the schema_checks
+
+        :return: The dictionary returned by the InstanceCore.check() method
+        :rtype: dict
+        :raises BadInstanceChecks: if the instance checks do not match the schema
+        :author: baobab soluciones
+        """
+        checks = self.check()
+        validator = Draft7Validator(self.schema_checks)
+        if not validator.is_valid(checks):
```
JaimeSotomayor, line 30: It would be interesting to include a warning message when, despite the schema validation being correct, it is detected that the tables or columns of the json being validated are not in the schema. And print the tables or columns that are not being checked. 
This applies to all schema validations.

ggsdc: Can you create an issue with this, because I do not think with the current jsonschema package taht we use is not possible but I will checkj it out, so I want to close this PR without this.

PR #580, opened 2024-12-04
Feature/socket

marioncottard:


marioncottard:
@ampuerin, in order for this to work, I had to replace gevent workers by GeventWebSocketWorkers, from the gevent-websocket library. I also added a monkey patch. None of those seem to be causing any problem in local with docker, but do you think those changes could have an influence on production deployments?

ggsdc:
> @ampuerin, in order for this to work, I had to replace gevent workers by GeventWebSocketWorkers, from the gevent-websocket library. I also added a monkey patch. None of those seem to be causing any problem in local with docker, but do you think those changes could have an influence on production deployments?

@ampuerin?

ggsdc:
Ok, to be able to use this in production we have to change the way we deploy so that the messages can be handled properly, @ampuerin it wuld be good if you could confirm this, but it is my understanding that:

1. to use flask-socketIO with gunicorn we have to use one worker per process (currently we use 3)
2. so if we want more workers we should run several instances behind a load balancer such as nginx.
3. use Redis to communicate between workers (can we reuse the Redis that Airflow uses?).

This is a more detailed explanation:

> ### Why You Can't Simply Use Multiple Workers with WebSockets
> The issue lies in how WebSocket connections work:
> 1. Connection Persistence: WebSocket connections are long-lived, persistent connections that maintain state.
> 2. Sticky Sessions Required: When a client establishes a WebSocket connection, all subsequent communications must be handled by the same server process that initially accepted the connection.
> 3. Load Balancing Limitations: Gunicorn's internal load balancer uses a simple round-robin algorithm that doesn't support "sticky sessions" needed for WebSockets. If you have 3 workers, a client might connect to worker # 1, but then the next message might go to worker # 2, which doesn't know about that connection.
> 4. Room Membership: In Socket.IO (which Flask-SocketIO implements), clients join "rooms" for targeted messaging. This room membership is stored in the worker's memory - if messages go to different workers, the room membership information is lost.
> As Miguel Grinberg (the author of Flask-SocketIO) explains in the Flask-SocketIO documentation:
> Due to the limited load balancing algorithm used by gunicorn, it is not possible to use more than one worker process when using this web server.
> ### How to Scale WebSockets with Multiple Processes
> If you need to scale beyond a single worker process (which is likely for production), you'll need a more sophisticated setup:
> 1. Multiple Single-Worker Instances: Run multiple gunicorn instances, each with a single worker (workers = 1) on different ports.
> 2. Proper Load Balancer: Use a load balancer like nginx in front, configured to support sticky sessions (usually based on IP or cookies).
> 3. Message Queue: Implement a message queue (like Redis or RabbitMQ) that allows the different worker processes to communicate and synchronize state.

PR #593, opened 2025-01-09
Enhance authentication and configuration handling

ggsdc:
- Added support for username and password login for service users in the OpenID authentication process, controlled by the new configuration option `SERVICE_USER_ALLOW_PASSWORD_LOGIN`.
- Introduced a new configuration class `TestingOpenAuth` for testing scenarios involving OpenID authentication.
- Updated the `LoginOpenAuthRequest` schema to validate that either a token is provided or both username and password are present, ensuring proper authentication flow.
- Enhanced the `LoginBaseEndpoint` to handle various authentication scenarios, including error handling for invalid requests.
- Improved documentation across configuration classes and schemas for better clarity and maintainability.

### DIFF
--- a/cornflow-server/cornflow/endpoints/login.py
+++ b/cornflow-server/cornflow/endpoints/login.py
@@ -181,6 +181,9 @@ class LoginBaseEndpoint(BaseMetaResource):
                 )
 
             username = decoded_token["preferred_username"]
+            email = decoded_token.get("email", f"{username}@test.org")
+            first_name = decoded_token.get("given_name", "")
+            last_name = decoded_token.get("family_name", "")
 
             user = self.data_model.get_one_object(username=username)
 
@@ -189,7 +192,12 @@ class LoginBaseEndpoint(BaseMetaResource):
                     f"OpenID user {username} does not exist and is created"
                 )
 
-                data = {"username": username, "email": username}
+                data = {
+                    "username": username,
+                    "email": email,
+                    "first_name": first_name,
+                    "last_name": last_name,
+                }
 
                 user = self.data_model(data=data)
                 user.save()
--- a/cornflow-server/cornflow/endpoints/login.py
+++ b/cornflow-server/cornflow/endpoints/login.py
@@ -38,6 +38,7 @@ class LoginBaseEndpoint(BaseMetaResource):
     def __init__(self):
         super().__init__()
         self.ldap_class = LDAPBase
+        self.user_role_association = UserRoleModel
 
     def log_in(self, **kwargs):
         """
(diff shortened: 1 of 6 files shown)


cornflow-server/cornflow/endpoints/login.py
```diff
@@ -136,58 +136,89 @@ def auth_ldap_authenticate(self, username, password):
...
+                data = {"username": username, "email": username}
 
-            user_role = self.user_role_association(
-                {"user_id": user.id, "role_id": int(current_app.config["DEFAULT_ROLE"])}
-            )
+                user = self.data_model(data=data)
+                user.save()
 
-            user_role.save()
+                self.user_role_association(user.id)
```
marioncottard, line 196: I am not sure that this line does something

cornflow-server/cornflow/schemas/user.py
```diff
@@ -66,9 +68,23 @@ class LoginEndpointRequest(Schema):
...
-    token = fields.Str(required=True)
+    token = fields.Str(required=False)
+    username = fields.Str(required=False)
+    password = fields.Str(required=False)
+
+    @validates_schema
+    def validate_fields(self, data, **kwargs):
+        if data.get("token") is None:
+            if not data.get("username") or not data.get("password"):
+                raise ValidationError(
```
JaimeSotomayor, line 82: This is not 100% true. It could be:
"A token or special user role needs to be provided when using Open ID authentication" or something similar.

cornflow-server/cornflow/tests/unit/test_log_in.py
```diff
@@ -29,5 +37,326 @@ def setUp(self):
...
+        Tests that a service user can not log in with username and password
+        """
+        response = self.client.post(
+            LOGIN_URL,
+            data=json.dumps(self.service_data),
+            headers={"Content-Type": "application/json"},
+        )
+
+        self.assertEqual(400, response.status_code)
+        self.assertEqual(response.json["error"], "Invalid request")
```
JaimeSotomayor, line 102: Should the "Invalid request" be a constant?

JaimeSotomayor (dismissed):

cornflow-server/cornflow/schemas/user.py
ggsdc: This error should only pop up if no token is available and either a username or a password is provided, a message and end-user is going to receive as the service user is not going to read them. So for an end-user I prefer to tell them that they need a token.

AlejandraGalan (approved):

cornflow-server/cornflow/tests/unit/test_log_in.py
ggsdc: Could be but I usually do not generalize those cases on tests.

ampuerin (approved):

marioncottard (approved):

PR #596, opened 2025-01-14
Updated action version

ggsdc:


ggsdc:
Bypass due to hotfix that only modifies publishing action

marioncottard (approved):

PR #597, opened 2025-01-15
Hotfix/publish action

ggsdc:


ggsdc:
Bypass due to only changes being on the github actions workflow (updated to altest recommendations by https://packaging.python.org/en/latest/ and the version on cornflow-client setup.py.

PR #602, opened 2025-02-04
Disable_detail method implemented in class BaseMetaResource

### LINKED ISSUE #599: Finish development of disable_detail method on BaseMetaResource
`BaseMetaResource` has a `disable_detail` method that is not implemented that could be of use when we do not want to directly delete information, just give it a `deleted_at` value.

To call the `disable` method on the `TraceAttributesModel`.

**Exaple of how the code could be**:

``` python

def disable_detail(self, idx):
  instance = self.data_model.query.get(idx)
  if row is None:
    raise ObjectNotFound()
  row.disable()
  return {"message": "Object marked as deleted"}, 200

```

Implementation should cover a use case: either with instances or executions and appropiate unit tests to check that behaviour is as expected.

Fidel-Rodriguez-Vidal:
Method implemented
fix #599

### DIFF
--- a/cornflow-server/cornflow/endpoints/meta_resource.py
+++ b/cornflow-server/cornflow/endpoints/meta_resource.py
@@ -13,9 +13,6 @@ from cornflow.shared.const import ALL_DEFAULT_ROLES
 from cornflow.shared.exceptions import InvalidUsage, ObjectDoesNotExist, NoPermission
 
 
-class ObjectNotFound(Exception):
-    pass
-
 
 class BaseMetaResource(Resource, MethodResource):
     """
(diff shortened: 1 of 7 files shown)


cornflow-server/cornflow/endpoints/meta_resource.py
```diff
@@ -13,6 +13,10 @@
 from cornflow.shared.exceptions import InvalidUsage, ObjectDoesNotExist, NoPermission
 
 
+class ObjectNotFound(Exception):
```
AlejandraGalan, line 16: We can use the predefined exception ObjectDoesNotExist instead of creating a new one.

AlejandraGalan:
Just a review PR. DO NOT MERGE.

AlejandraGalan (approved):

ggsdc (approved):

PR #603, opened 2025-02-04
Feature/issue568

### LINKED ISSUE #568: Use an instance with data={} for execution.
**Describe the bug**
When attempting to use an instance initialized with an empty dictionary (data={}), an error occurs. This error arises because the instance's table in the database incorrectly stores the empty dictionary as null rather than {}, leading to execution issues.

**To Reproduce**
1. Create an instance with data={}.
2. Attempt to execute the instance.
3. Observe the error:
`{
    "jsonschema_errors": [
        "None is not of type 'object'\n\nFailed validating 'type' in schema:\n    {'$schema': 'http://json-schema.org/draft-07/schema#',\n     'properties': {},\n     'required': [],\n     'type': 'object'}\n\nOn instance:\n    None"
    ],
    "error": "The data sent is not valid"
`

**Expected behavior**
We sould be saving an empty dict in the database instead of null so it works correctly

**Metadata (please complete the following information):**
 - cornflow version 1.1.2

alejandrarr1601:
Closes #568

### DIFF
--- a/cornflow-server/cornflow/tests/data/empty_instance.json
+++ b/cornflow-server/cornflow/tests/data/empty_instance.json
@@ -1,0 +1,5 @@
+{"data": {
+},
+"name":"instance_1",
+"description":"This is a test instance to check if name and description get saved succesfully",
+"schema":  "gc"}
--- a/cornflow-server/cornflow/tests/data/empty_instance.json
+++ b/cornflow-server/cornflow/tests/data/empty_instance.json
@@ -1,5 +1,7 @@
-{"data": {
-},
-"name":"instance_1",
-"description":"This is a test instance to check if empty dictionary get saved succesfully",
-"schema":  "gc"}
+{
+  "data": {
+  },
+  "name": "instance_1",
+  "description": "This is a test instance to check if empty dictionary get saved succesfully",
+  "schema": "gc"
+}
--- a/cornflow-server/cornflow/tests/data/empty_instance.json
+++ b/cornflow-server/cornflow/tests/data/empty_instance.json
@@ -1,5 +1,5 @@
 {"data": {
 },
 "name":"instance_1",
-"description":"This is a test instance to check if name and description get saved succesfully",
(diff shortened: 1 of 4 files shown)


cornflow-server/cornflow/tests/unit/test_instances.py
```diff
@@ -49,6 +51,12 @@ def load_file(_file):
     def test_new_instance(self):
         self.create_new_row(self.url, self.model, self.payload)
 
+    def test_empty_instance(self):
+        """
+        testing what happend with empty data
```
AlejandraGalan, line 56: We could have a more specific docstring. I would describe in more detail the problem we are testing to ensure it does not occur.

cornflow-server/cornflow/tests/unit/test_instances.py
```diff
@@ -49,6 +51,12 @@ def load_file(_file):
     def test_new_instance(self):
         self.create_new_row(self.url, self.model, self.payload)
 
+    def test_empty_instance(self):
+        """
+        testing what happend with empty data 
+        """
+        self.create_new_row(self.url, self.model, self.payload2)
+
```
AlejandraGalan, line 59: We could add a get and check that we find an empty dict where expected. We can use something like this to achieve that:
```
rows = self.client.get(
            url, follow_redirects=True, headers=self.get_header_with_auth(token)
        )
```

AlejandraGalan (commented):
Some small changes

cornflow-server/cornflow/tests/unit/test_instances.py
ggsdc: Is this comment solved?

cornflow-server/cornflow/tests/data/empty_instance.json
ggsdc, line 1: Could we format this file a little bit better?

ggsdc (approved):

PR #605, opened 2025-02-13
Feature/issue 569

### LINKED ISSUE #569: Fix get_solver_config method from Experiment
**Describe the bug**
The get_solver_config method is returning every key in the config dict as solver parameters even when solver_config is defined.

**To Reproduce**
Pass the following config dict in the execution of any app:
```
config = dict(
        solver="milp_solver.gurobi",
        msg=True,
        solver_config=dict(time_limit=10 * 60, rel_gap=0.001, MIPGapAbs=5),
)
```

Use the following code in the experiment (options inf experiment is config in the app.solve):
```
solver_config = self.get_solver_config(options)
opt = SolverFactory(solver_name)
opt.options.update(solver_config)
results = opt.solve(model_instance, tee=options.get("msg", True))
```
solver_config will include "msg" which is not an option of the solver and may generate an error when passed to the solver.
Also the description of the function states that it will return the name of the solver and it doesn't.
The option remove_unknown cannot be used in this case because it would also remove any solver option which is not in the mapping (like MIPGapAbs here)

The way around I am currently using is this code:
`opt.options.update(self.get_solver_config(options["solver_config"], default_solver=solver_name))`

**Expected behavior**
This method should only return the data in solver_config if it is present or it should at least have an option to do it that way.

I can do a PR for it if we agree on how this method should behave.

alejandrarr1601:
Review issue 569
fix #569

### DIFF
--- a/libs/client/cornflow_client/core/experiment.py
+++ b/libs/client/cornflow_client/core/experiment.py
@@ -128,7 +128,7 @@ class ExperimentCore(ABC):
         """
         Format the configuration used to solve the problem.
         Solver configuration can either be directly in config using cornflow mapping name
-           or in a config["solver_config"] using the solver names.
+           or in a config["solver_config"] using the solver names. Nothing is ever removed from solver_config.
         Example:
             config = {
                 "solver":"milp.cbc",
--- a/libs/client/cornflow_client/core/experiment.py
+++ b/libs/client/cornflow_client/core/experiment.py
@@ -168,7 +168,7 @@ class ExperimentCore(ABC):
             if config.get("solver_config"):
                 conf.update(
                     {
-                        mapping[k, lib, solver]: v
+                        mapping.get((k, lib, solver), k): v
                         for k, v in config["solver_config"].items()
                     }
                 )
--- a/libs/client/cornflow_client/core/experiment.py
+++ b/libs/client/cornflow_client/core/experiment.py
@@ -170,7 +170,6 @@ class ExperimentCore(ABC):
                     {
                         mapping[k, lib, solver]: v
                         for k, v in config["solver_config"].items()
-                        if (k, lib, solver) in mapping
                     }
                 )
         else:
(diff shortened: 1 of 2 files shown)


AlejandraGalan (approved):

AlejandraGalan:
I would like to know your feedback @hugolarzabal @ggsdc

libs/client/cornflow_client/core/experiment.py
```diff
@@ -168,9 +168,8 @@ def get_solver_config(
             if config.get("solver_config"):
                 conf.update(
                     {
-                        mapping[k, lib, solver]: v
+                        mapping.get((k, lib, solver), k): v
                         for k, v in config["solver_config"].items()
```
hugolarzabal, line 172: the docstring of the method should also be edited to state that nothing is ever removed from solver_config

libs/client/cornflow_client/tests/unit/test_instance_solution_methods.py
```diff
@@ -386,6 +386,58 @@ def test_get_solver_config_pyomo(self):
...
 
+    def test_get_solver_config_with_remove_unknow(self):
+        config = dict(
+            solver="milp_solver.gurobi",
+            msg=True,
+            abs_gap=5,
+            solver_config=dict(time_limit=10 * 60, rel_gap=0.001,  additional = True)
+        )
+        res = self.class_to_use.get_solver_config(config, remove_unknown= True )
+        self.assertEqual(res, {"TimeLimit": 600, "MIPGap": 0.001 ,"MIPGapAbs": 5, "additional":True})
```
hugolarzabal, line 397: the arguments of assertEqual are Expected, Actual in that order.

This is not a big deal, but if they are in the opposite order, and the test fail, the error message might be confusing.

hugolarzabal: also, I think a cleaner way to write the test is:
```
...
expected = {"TimeLimit": 600, "MIPGap": 0.001 ,"MIPGapAbs": 5, "additional":True}
result = self.class_to_use.get_solver_config(config, remove_unknown= True )
msg="if remove_unknown is true, unmapped argument should be removed from config but not from solver_config"
self.assertEqual(expected, result, msg=msg)
```

hugolarzabal (changes requested):

hugolarzabal (approved):

ampuerin (dismissed):

ggsdc (approved):

ggsdc (approved):

PR #607, opened 2025-02-18
Feature/some doc

ggsdc:
This should make it so that if we want to define some extra permissions for the default roles on the external apps we can do so on the const file for the external app (if it is on the shared module) and that way we do not need custom commands for the permissions or the init service.

This PR also brings all changes on the develop branch of cornflow to the databricks feature branch


ggsdc:
The main files to review are:

- cli/service.py
- commands/permissions.py

PR #608, opened 2025-02-18
Changes to allow external apps to define extra permissions on default…

ggsdc:
… roles on the endpoints from the external app.

### DIFF
--- a/libs/client/cornflow_client/tests/integration/test_raw_cornflow_integration.py
+++ b/libs/client/cornflow_client/tests/integration/test_raw_cornflow_integration.py
@@ -76,7 +76,7 @@ class TestRawCornflowClientUser(TestCase):
         self.assertEqual(response.status_code, 200)
         data = response.json()
         self.assertEqual(data["cornflow_status"], "healthy")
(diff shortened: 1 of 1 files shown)


ampuerin (approved):

AlejandraGalan (approved):

cornflow-server/cornflow/shared/databricks.py
```diff
@@ -32,80 +39,91 @@ def from_config(cls, config):
...
         #   Entender cómo se usa checks_only
-        payload = dict(job_id=orch_name, notebook_parameters=dict(checks_only=checks_only))
+        payload = dict(
+            job_id=orch_name, notebook_parameters=dict(checks_only=checks_only)
+        )
         return self.request_headers_auth(method="POST", url=url, json=payload)
-    
-    def get_run_status(self, run_id):
+
+    def get_run_status(self, schema, run_id):
```
AlejandraGalan, line 99: This class is duplicated on client, so it should be the same on both

marioncottard (approved):

PR #629, opened 2025-04-08
Feature/databricks

AlejandraGalan:


cornflow-server/cornflow/cli/service.py
```diff
@@ -70,6 +72,7 @@ def init_cornflow_service():
                 config["cornflow_service_email"],
                 config["cornflow_service_pwd"],
             )
+        if config["cornflow_backend"] == AIRFLOW_BACKEND:
             _sync_with_airflow(
```
marioncottard, line 76: I think this block needs to be inside the "with app.app_context()" block

cornflow-server/cornflow/schemas/execution.py
```diff
@@ -78,6 +78,7 @@ class ExecutionEditRequest(Schema):
     name = fields.Str()
     description = fields.Str()
     data = fields.Raw()
+    instance_id = fields.Str(required=False)
```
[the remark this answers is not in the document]
AlejandraGalan: We need the flexibility for Nippon's project since we create the instance in the orchestrator instead of creating it making use in cornflow.

libs/client/cornflow_client/airflow/api.py
```diff
@@ -61,31 +67,35 @@ def set_dag_run_state(self, dag_name, payload):
...
         return self.consume_dag_run(dag_name, payload={}, method="POST")
 
     def update_dag_registry(self, dag_name="update_dag_registry"):
         return self.consume_dag_run(dag_name, payload={}, method="POST")
 
-    def get_dag_run_status(self, dag_name, dag_run_id):
+    def get_run_status(self, schema, run_id):
         return self.consume_dag_run(
-            dag_name, payload=None, dag_run_id=dag_run_id, method="GET"
+            schema, payload=None, dag_run_id=run_id, method="GET"
```
ggsdc, line 91: Same for this one

libs/client/cornflow_client/airflow/api.py
```diff
@@ -105,9 +115,12 @@ def set_dag_run_to_fail(self, dag_name, dag_run_id, new_status="failed"):
...
 
-    def get_dag_info(self, dag_name, method="GET"):
-        url = f"{self.url}/dags/{dag_name}"
-        return self.request_headers_auth(method=method, url=url)
+    def get_orch_info(self, orch_name, method="GET"):
+        url = f"{self.url}/dags/{orch_name}"
+        schema_info = self.request_headers_auth(method=method, url=url)
+        if schema_info.status_code != 200:
+            raise AirflowError("DAG not available")
+        return schema_info
```
ggsdc, line 123: `get_workflow_info`

cornflow-server/cornflow/endpoints/execution.py
```diff
@@ -282,6 +346,50 @@ def __init__(self):
...
+
+    @property
+    def orch_to_state_map(self):
+        self._init_orch()
+        return self._orch_to_state_map
+
+    @property
+    def orch_const(self):
+        self._init_orch()
+        return self._orch_const
```
ggsdc, line 392: As this properties get repeated. Does it make sense to have a base class that all endpoints inherit from so this can only be set up once?

libs/client/cornflow_client/airflow/api.py
AlejandraGalan: This functionality is not implemented for any other orchestrators, so for now, I would keep the original name so we can distinguish between the two cases.

cornflow-server/cornflow/commands/dag.py
```diff
@@ -40,7 +40,7 @@ def register_deployed_dags_command(
     }
 
     processed_dags = [
```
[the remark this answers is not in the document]
AlejandraGalan: This code only applies to the Airflow context.

cornflow-server/cornflow/endpoints/execution.py
```diff
@@ -488,23 +639,27 @@ def delete(self, idx):
     @authenticate(auth_class=Auth())
     @Auth.dag_permission_required
     def post(self, idx):
+        if self.orch_type != AIRFLOW_BACKEND:
+            return {
+                "message": f"This feature is not available for {self.orch_const['name']}"
+            }, 200
```
marioncottard, line 645: This should probably raise an exception or return an error status code

cornflow-server/cornflow/shared/constants_cli.py
marioncottard, line 1: Can this be deleted?

ggsdc (dismissed):

PR #641, opened 2025-05-23
Exposes the version of the Git tag at the /health endpoint.

DanielCruzVillameriel:


ggsdc:
I am going to close it for now as I do not like the approach of writing on a file during the github workflow.

Issue #620, opened 2025-04-02
Cornflow_client needs to have less restrictive dependencies


pchtsp:
**Describe the bug**
When installing cornflow_client, I cannot use "recent" versions of pulp or pytups. Because it has upper limits to their version numbers.

**To Reproduce**
pip install cornflow_client pytups pulp

**Expected behavior**
Version 1.0.3 of pytups is installed.
Version 3.1.1 of pulp is installed.

**Additional context**
Maybe there could be a < for major versions, such as:

pytups<2
pulp<4
ortools<10

Alternatively, and I think it's the correct way, the versions for cornflow_client should be set free and it should be cornflow that has strict versions of depedencies.

Not mandatory, but for cornflow dependencies, this could be achieved by switching to `pyproject.toml` and `uv.lock` instead of `requirements.txt`

I can make a PR if it's needed

PR #644, opened 2025-05-30
free library versions from cornflow_client

pchtsp:
it seems to pass all the unit tests locally. Fixes https://github.com/baobabsoluciones/cornflow/issues/620


ggsdc:
I'm going to close this PR, but not because I do not want to free up the versions. I am going to do it. But I am going to do it on cornflow-server at the same time, using a [cosntraints](https://pip.pypa.io/en/stable/user_guide/#constraints-files) file to handle that the versions of librares actually installed on our cornflow images do stay consistent with what has been tested, but allows to free the versions on local development.

This change will come as well changing both modules to use pyproject.toml instead of the old setup.py.

If you want to make all the cahnges yourself, you are welcome to do so and reopen the PR with all the changes, but please tell me before I start the development myself (sometime in the following two weeks)

pchtsp:
I will do a proposal using pyproject.toml + uv.lock (which I believe removes the need for a constraints file).

PR #648, opened 2025-06-05
All changes for permissions

AlejandraGalan:


cornflow-server/cornflow/commands/permissions.py
```diff
@@ -4,24 +4,43 @@
...
 
 
 def register_base_permissions_command(external_app: str = None, verbose: bool = False):
     if external_app is None:
-        from cornflow.endpoints import resources, alarms_resources
         resources_to_register = resources
+        extra_permissions = EXTRA_PERMISSION_ASSIGNATION
         if current_app.config["ALARMS_ENDPOINTS"]:
             resources_to_register = resources + alarms_resources
     elif external_app is not None:
```
hugolarzabal, line 24: this should be an else

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -4,24 +4,43 @@
...
+            extra_permissions = EXTRA_PERMISSION_ASSIGNATION
+
+        if current_app.config["ALARMS_ENDPOINTS"]:
+            resources_to_register = (
+                external_module.endpoints.resources + resources + alarms_resources
+            )
+        else:
+            resources_to_register = external_module.endpoints.resources + resources
+
     else:
```
hugolarzabal, line 42: an else for unreachable code doesn't make sense

cornflow-server/cornflow/cli/service.py
```diff
@@ -89,9 +105,27 @@ def init_cornflow_service():
...
         external_app_lib = import_module(external_app_module)
         app = external_app_lib.create_wsgi_app(environment, cornflow_db_conn)
-
+        logger.info(
+            "************************ EXTERNAL APP CREATED SUCCESSFULLY ************************"
+        )
+        click.echo("------------------------Initializing database----------------")
+        click.echo(app)
+        logger.info(
+            "************************ APP INSTANCE: {type(app)} ************************"
```
marioncottard, line 114: I think you're missing the `f` for this f-string

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -74,15 +135,64 @@ def register_base_permissions_command(external_app: str = None, verbose: bool =
...
+            db.session.rollback()
+            current_app.logger.error(
+                f"Integrity error on base permissions register: {e}"
+            )
+        except DBAPIError as e:
+            db.session.rollback()
+            current_app.logger.error(f"Unknown error on base permissions register: {e}")
+
+    # We check the roles have been correctly registered
+    roles_in_db = [role.id for role in RoleModel.get_all_objects()]
```
hugolarzabal, line 153: does it really make sense to check the previous action?

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -4,24 +4,43 @@
...
+
+        if current_app.config["ALARMS_ENDPOINTS"]:
+            resources_to_register = (
+                external_module.endpoints.resources + resources + alarms_resources
+            )
+        else:
+            resources_to_register = external_module.endpoints.resources + resources
+
     else:
+        # Not reachable code
```
marioncottard, line 43: If it's not reachable, it should probably be deleted, and the "elif" replaced with an "else"

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -74,15 +135,64 @@ def register_base_permissions_command(external_app: str = None, verbose: bool =
...
+        # Get permissions that were filtered out because they already exist
+        existing_permissions = [
+            permission
+            for permission in permissions_to_register
+            if (permission.role_id, permission.action_id, permission.api_view_id)
+            in permissions_in_db_keys
+        ]
+
+        if existing_permissions:
+            for perm in existing_permissions:
```
hugolarzabal, line 186: that might be too much log

hugolarzabal: maybe just log: "x permission existing, y new permissions"

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -31,28 +50,70 @@ def register_base_permissions_command(external_app: str = None, verbose: bool =
...
+    new_roles_to_add = []
+    if len(additional_roles_with_access) > 0:
+        # If our id is not in the existing roles, we add it, applying a
+        # pre-defined name custom_role_<id>
+        for custom_role in additional_roles_with_access:
+            if custom_role not in existing_roles:
+                new_role = RoleModel(
+                    {
+                        "id": custom_role,
+                        "name": f"custom_role_{custom_role}",
```
marioncottard, line 79: For now I think this is fine, but in the long run we probably should find a way to get the names and IDs of the extra roles directly from the external app

cornflow-server/cornflow/commands/views.py
```diff
@@ -68,3 +41,108 @@ def register_views_command(external_app: str = None, verbose: bool = False):
             current_app.logger.info("No new endpoints to be registered")
 
     return True
+
+
+def load_changes_to_db(views_to_delete, views_to_register, views_to_update):
```
hugolarzabal, line 46: missing doc strings

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -31,28 +50,70 @@ def register_base_permissions_command(external_app: str = None, verbose: bool =
...
+                    }
+                )
+                new_roles_to_add.append(new_role)
+
+        # Create extended permission assignation including additional roles
+        base_permissions_assignation = BASE_PERMISSION_ASSIGNATION + [
+            (custom_role, action)
+            for custom_role in additional_roles_with_access
+            for action in ALL_DEFAULT_ACTIONS
```
marioncottard, line 88: Same here, I think we should have some way of defining from the external app what actions each custom role can perform

marioncottard: Because (correct me if I'm wrong) it looks like you're actually giving permission to these roles to perform all actions on the endpoints it has access to? Which can be dangerous too

cornflow-server/cornflow/commands/views.py
```diff
@@ -68,3 +41,108 @@ def register_views_command(external_app: str = None, verbose: bool = False):
...
+        db.session.commit()
+    except IntegrityError as e:
+        db.session.rollback()
+        current_app.logger.error(f"Integrity error on views register: {e}")
+    except DBAPIError as e:
+        db.session.rollback()
+        current_app.logger.error(f"Unknow error on views register: {e}")
+
+
+def get_views_to_update_and_delete(
```
hugolarzabal, line 66: I think it might be better to just make a different function for viers_to update and views_to_delete.

Also views_to_delete code is pretty easy:
[view_attrs["id"] for view_name, view_attrs in views_registered_urls_all_attributes.items() if view_name not in all_resources_to_register_views_endpoints.keys()

hugolarzabal: and both arguments could have a simpler name

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -74,15 +135,64 @@ def register_base_permissions_command(external_app: str = None, verbose: bool =
...
-    # if len(permissions_to_delete) > 0:
-    #     for permission in permissions_to_delete:
-    #         db.session.delete(permission)
+        # Filter out permissions that already exist in the database
+        permissions_to_register = [
+            permission
+            for permission in permissions_to_register
+            if (permission.role_id, permission.action_id, permission.api_view_id)
+            not in permissions_in_db_keys
+        ]
```
marioncottard, line 176: Isn't this filter the same as the one above?

cornflow-server/cornflow/commands/views.py
```diff
@@ -68,3 +41,108 @@ def register_views_command(external_app: str = None, verbose: bool = False):
...
+        sys.path.append("./")
+        external_module = import_module(external_app)
+        if current_app.config["ALARMS_ENDPOINTS"]:
+            resources_to_register = (
+                external_module.endpoints.resources + resources + alarms_resources
+            )
+        else:
+            resources_to_register = external_module.endpoints.resources + resources
+    else:
+        # Not reachable code
```
marioncottard, line 144: If this is not reachable, it should be deleted

cornflow-server/cornflow/commands/views.py
```diff
@@ -68,3 +41,108 @@ def register_views_command(external_app: str = None, verbose: bool = False):
...
+                )
+        else:
+            views_to_delete.append(view_attrs["id"])
+    return views_to_delete, views_to_update
+
+
+def get_views_to_register(resources_to_register, views_registered):
+    views_to_register = [
+        ViewModel(
+            {
```
hugolarzabal, line 95: I think that you should just create the dict here and not apply view_model.

And just return views_to_register in the function

cornflow-server/cornflow/commands/views.py
```diff
@@ -68,3 +41,108 @@ def register_views_command(external_app: str = None, verbose: bool = False):
...
+                "name": view["endpoint"],
+                "url_rule": view["urls"],
+                "description": view["resource"].DESCRIPTION,
+            }
+        )
+        for view in resources_to_register
+        if view["endpoint"] not in views_registered
+    ]
+    all_resources_to_register_views_endpoints = {
```
hugolarzabal, line 104: this should be formatted in the function which use it

cornflow-server/cornflow/commands/views.py
```diff
@@ -68,3 +41,108 @@ def register_views_command(external_app: str = None, verbose: bool = False):
...
+            "description": view["resource"].DESCRIPTION,
+        }
+        for view in resources_to_register
+    }
+    return all_resources_to_register_views_endpoints, views_to_register
+
+
+def get_database_view():
+    views_registered = [view.name for view in ViewModel.get_all_objects()]
+    views_registered_urls_all_attributes = {
```
hugolarzabal, line 116: you could return only this dict.
the views_registered list is just the keys from this dict

cornflow-server/cornflow/commands/views.py
```diff
@@ -68,3 +41,108 @@ def register_views_command(external_app: str = None, verbose: bool = False):
...
+        current_app.logger.info(f" USING EXTERNAL APP: {external_app} ")
+        sys.path.append("./")
+        external_module = import_module(external_app)
+        if current_app.config["ALARMS_ENDPOINTS"]:
+            resources_to_register = (
+                external_module.endpoints.resources + resources + alarms_resources
+            )
+        else:
+            resources_to_register = external_module.endpoints.resources + resources
+    else:
```
hugolarzabal, line 143: do not use elif when the elif is actually all other cases. This why else exist.

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -4,24 +4,43 @@
...
-from cornflow.models import ViewModel, PermissionViewRoleModel
+from cornflow.models import ViewModel, PermissionViewRoleModel, RoleModel
 from cornflow.shared import db
+
 from flask import current_app
 from sqlalchemy.exc import DBAPIError, IntegrityError
+from cornflow.endpoints import resources, alarms_resources
 
 
 def register_base_permissions_command(external_app: str = None, verbose: bool = False):
```
hugolarzabal, line 18: this function could also be separated in smaller functions

hugolarzabal (changes requested):

cornflow-server/cornflow/commands/permissions.py
AlejandraGalan: I agree. Maybe we should just assign the lowest permissions by default and allow adding new ones using the extra permissions constant in the external app. How does that sound?

cornflow-server/cornflow/commands/permissions.py
AlejandraGalan: I might delete this log, since it was mainly used during the development of the fix. Would you prefer to keep it?

cornflow-server/cornflow/commands/permissions.py
AlejandraGalan: it was mainly used during the development of the fix, i will delete it

cornflow-server/cornflow/commands/permissions.py
hugolarzabal: I think telling the number of new permissions is a good way for the user to see if new endpoints or roles are taken into account

cornflow-server/cornflow/commands/views.py
hugolarzabal: for loops with append are less efficient than list comprehensions, so separating it might actually be faster

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -4,85 +4,100 @@
...
-            {
-                "role_id": role,
-                "action_id": action,
-                "api_view_id": views_in_db[endpoint],
-            }
-        )
-        for role, action, endpoint in EXTRA_PERMISSION_ASSIGNATION
-    ]
+    if verbose:
+        if len(permissions_to_register) > 0:
```
hugolarzabal, line 57: I think that this one of the msot important logs so it may be always activated

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -163,13 +158,18 @@ def get_permissions_in_code_as_tuples(
...
     permissions_tuples = set()
 
     # Add permissions from ROLES_WITH_ACCESS
     for role, action in base_permissions_assignation:
         for view in resources_to_register:
             if role in view["resource"].ROLES_WITH_ACCESS:
+                # For custom roles (not in ALL_DEFAULT_ROLES), only grant GET access
+                if role not in ALL_DEFAULT_ROLES and action != GET_ACTION:
+                    continue
                 permissions_tuples.add((role, action, views_in_db[view["endpoint"]]))
```
marioncottard, line 173: Is there a reason why you have this `if` here instead of having this in get_base_permission():
```
base_permissions_assignation = BASE_PERMISSION_ASSIGNATION + [
        (custom_role, GET_ACTION)
        for custom_role in all_custom_roles_in_access
]
```
instead of this:
```
base_permissions_assignation = BASE_PERMISSION_ASSIGNATION + [
        (custom_role, action)
        for custom_role in all_custom_roles_in_access
        for action in ALL_DEFAULT_ACTIONS
]
```
?

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -163,13 +158,18 @@ def get_permissions_in_code_as_tuples(
     """
     Get the permissions in code as tuples.
     """
+    from cornflow.shared.const import ALL_DEFAULT_ROLES, GET_ACTION
+
```
marioncottard, line 162: This import should be at the top of the file

cornflow-server/cornflow/commands/permissions.py
```diff
@@ -105,25 +111,108 @@ def register_base_permissions_command(external_app: str = None, verbose: bool =
...
+        ]
+    )
+
+    # Create extended permission assignation including all custom roles
+    # For custom roles (not in ALL_DEFAULT_ROLES), only grant GET access
+    base_permissions_assignation = BASE_PERMISSION_ASSIGNATION + [  
+        (custom_role, GET_ACTION)  
+        for custom_role in all_custom_roles_in_access  
+]  
+
```
marioncottard, line 192: Do you have black activated?

marioncottard (approved):

Issue #532, opened 2024-05-11
Add .dagignore file


ggsdc:
Current update all schemas dag has a list of files that need to be ignored in order to better process the active apps on Airflow.
This list could come from a .dagignore file and if not present default to the current list.

Probably, it would be better if just in case the .dagignore file is extended with the current list

PR #649, opened 2025-06-06
Feature/issue 532

alejandrarr1601:
issue #532


cornflow-dags/DAG/update_all_schemas.py
```diff
@@ -34,30 +33,67 @@ def get_new_apps() -> List[ApplicationCore]:
...
+
+    if os.path.exists(dagignore_path):
+        try:
+            with open(dagignore_path, 'r') as f:
+                for line in f:
+                    line = line.strip()
+                    if line and not line.startswith('#'):
+                        ignore_patterns.add(line)
+            print(f"Loaded ignore patterns from {dagignore_path}")
+        except Exception as e:
```
ggsdc, line 62: I guess the error here should be something like `FileNotFoundError`, no? I would prefer if the only expcetion that we capture is that one.

ggsdc (changes requested):
Looks good to me, but can we make the small change that I request?

cornflow-dags/DAG/update_all_schemas.py
```diff
@@ -53,14 +46,14 @@ def _load_ignore_patterns():
...
     if os.path.exists(dagignore_path):
         try:
-            with open(dagignore_path, 'r') as f:
+            with open(dagignore_path, "r") as f:
                 for line in f:
                     line = line.strip()
-                    if line and not line.startswith('#'):
+                    if line and not line.startswith("#"):
                         ignore_patterns.add(line)
             print(f"Loaded ignore patterns from {dagignore_path}")
```
AlejandraGalan, line 54: I know its not from this PR, but maybe we should replace these prints by logs.

PR #706, opened 2026-02-05
Feature/try fix dependabot

AlejandraGalan:
With this configuration file, we ensure client requirements are analyzed. We should fix all requirements and avoid ranges for it to work properly .


AlejandraGalan:
Check changes needed to merge with UV changes

PR #719, opened 2026-04-16
Issue permissions dag 

alejandrarr1601:


cornflow-server/cornflow/endpoints/dag.py
```diff
@@ -262,7 +262,13 @@ def get(self):
     @marshal_with(DeployedDAGSchema)
     @use_kwargs(DeployedDAGSchema)
     def post(self, **kwargs):
-        return self.post_list(kwargs)
+        response = self.post_list(kwargs)
+        # New workflows must get permission rows for existing users (e.g. admin created
+        # at init before DAGs existed). Same logic as register_dag_permissions on boot.
+        from cornflow.commands.permissions import register_dag_permissions_command
+
+        register_dag_permissions_command(verbose=False)
```
marioncottard, line 270: Two things: 
- Can we move this import to the top of the file?
- It would probably be cleaner to register the dag permissions only for the dag that was just created. You could add a "dag_id" argument to register_dag_permissions_command, and change register_dag_permissions_command so that if this argument is not provided, it registers the permissions for all dags, and if it is provided, it registers the permissions only for the provided dag.

PR #720, opened 2026-04-16
Add get role for other roles

AlejandraGalan:


cornflow-server/cornflow/tests/unit/test_roles.py
```diff
@@ -253,11 +254,37 @@ def test_get_user_roles_authorized_user(self):
...
+            )
+            self.assertEqual(200, response.status_code)
+            self.assertTrue(
+                any(
+                    r["user_id"] == user.id and r["role_id"] == role
+                    for r in response.json
+                )
+            )
 
     def test_get_user_roles_not_authorized_user(self):
```
marioncottard: If all roles are now allowed to access this endpoint, and assuming that all users have at least one role, I don't think this test is needed anymore.

marioncottard (dismissed):

marioncottard (approved):

PR #722, opened 2026-05-08
Feature/execution files

marioncottard:


cornflow-dags/requirements.txt
```diff
@@ -1,7 +1,7 @@
-cornflow-client==1.3.5rc3
+cornflow-client @ git+https://github.com/baobabsoluciones/cornflow@feature/execution_files#subdirectory=libs/client
```
AlejandraGalan: Needs to be changed

cornflow-dags/requirements.txt
```diff
@@ -1,7 +1,7 @@
...
 orloge==0.17.2
 PuLP==2.9.0
-pandas==2.2.0
+pandas>=2.3.3,<3.0.0
```
AlejandraGalan: We should avoid ranges in requirements, is it needed?

libs/client/requirements.txt
```diff
@@ -2,6 +2,8 @@ requests==2.32.4
 genson==1.2.2
 jsonschema==4.21.1
 marshmallow==3.26.2
+openpyxl>=3.1.2,<4.0.0
```
AlejandraGalan: We should avoid ranges

AlejandraGalan (approved):

PR #728, opened 2026-06-10
Security: require admin/service passwords instead of hardcoded defaults

ggsdc:
## Problem
`cornflow/cli/service.py` fell back to hardcoded passwords for the bootstrap admin and service users when the env vars were not set:

```python
cornflow_admin_pwd = os.getenv("CORNFLOW_ADMIN_PWD", "Cornflow_admin1234")
cornflow_service_pwd = os.getenv("CORNFLOW_SERVICE_PWD", "Service_user1234")
```

Combined with `SERVICE_USER_ALLOW_PASSWORD_LOGIN=1` (default), any operator who forgot to set these env vars would boot a server with publicly-known admin credentials — trivial full compromise of an exposed API.

## Fix
The service now refuses to start unless `CORNFLOW_ADMIN_PWD` and `CORNFLOW_SERVICE_PWD` are explicitly provided, exiting with a clear `FATAL` message otherwise. No default passwords remain.

## Impact / migration
Deployments that relied on the implicit defaults must now set `CORNFLOW_ADMIN_PWD` and `CORNFLOW_SERVICE_PWD`. This is intentional hardening.

### DIFF
--- a/cornflow-server/cornflow/cli/service.py
+++ b/cornflow-server/cornflow/cli/service.py
@@ -197,12 +197,22 @@ def _setup_environment_variables():
     cornflow_admin_email = os.getenv(
         "CORNFLOW_ADMIN_EMAIL", "cornflow_admin@cornflow.com"
     )
-    cornflow_admin_pwd = os.getenv("CORNFLOW_ADMIN_PWD", "Cornflow_admin1234")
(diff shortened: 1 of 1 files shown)


AlejandraGalan (approved):

AlejandraGalan:
We should allow a default value for local deployments. We should be able to differentiate using the flask environment variable.

PR #732, opened 2026-06-10
Security: disable password-recovery endpoint by default (interim mitigation)

ggsdc:
## Problem
`RecoverPassword` (`/user/recover-password/`) immediately generates a new random password and **overwrites the target user's password** on an unauthenticated request, with no rate limiting. Anyone who knows a user's email can lock that user out at will (account-lockout DoS) and mail-bomb them.

## Fix (interim, smallest change)
Gate the endpoint behind a new `PASSWORD_RECOVERY_ENABLED` config flag, **default off**. When disabled the endpoint returns `501`/not-implemented and changes nothing. This removes the vulnerability from default deployments without a schema change.

## Follow-up
The proper fix — a token-based reset flow (email a time-limited link, confirm before changing the password) — is tracked in a separate issue and will supersede this flag.

## Impact / migration
Deployments that intentionally use the current recovery behaviour must set `PASSWORD_RECOVERY_ENABLED=1`.

### DIFF
--- a/cornflow-server/cornflow/config.py
+++ b/cornflow-server/cornflow/config.py
@@ -102,6 +102,11 @@ class DefaultConfig(object):
     # compress config
     COMPRESS_REGISTER = False
 
+    # Password recovery endpoint. Disabled by default: when enabled, the current
+    # endpoint resets the user's password on an unauthenticated request, which
+    # allows account-lockout abuse. Keep off until the token-based flow lands.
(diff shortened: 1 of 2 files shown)


AlejandraGalan:
I don't think this is the right approach. Password regeneration should be handled by the Service Admin. Alternatively, we could send an email with the necessary instructions so the user can change it themselves, ensuring proper ownership. @ggsdc

PR #758, opened 2026-07-11
Route DAG/instance/execution/case data endpoints to raw-JSON variants

ggsdc:
## Summary

Endpoints that return an execution's/instance's/case's `data` field (and siblings) suffered performance issues when that JSON field is large, because the normal path decodes the stored JSON into Python objects and then re-encodes it back to JSON for the HTTP response — pure wasted CPU for fields that are returned verbatim.

This PR:
- Adds "raw" variants for the four affected endpoints that cast the JSON column to text at the SQL layer (skipping decode) and splice the already-valid JSON text directly into a hand-built response body (skipping re-encode): `DAGDetailEndpointRaw`, `InstanceDataEndpointRaw`, `ExecutionDataEndpointRaw`, `CaseDataEndpointRaw`.
- Routes them in place of the original implementations:
  - `GET /dag/<idx>/`
  - `GET /instance/<idx>/data/`
  - `GET /execution/<idx>/data/`
  - `GET /case/<idx>/data/`
- Keeps the original classes (`DAGDetailEndpoint`, `DAGDetailEndpointFast`, `InstanceDataEndpoint`, `ExecutionDataEndpoint`, `CaseDataEndpoint`) in the codebase, marked **DEPRECATED** and **no longer routed**. They're retained because the `*Raw` classes inherit their non-GET methods (`put`/`patch`/`delete`) from them, and so the benchmark scripts can still exercise them for comparison.
- Adds standalone benchmark scripts (`cornflow/tests/load/`, not collected by pytest) that build ~100MB JSON fixtures and time the deprecated vs. live implementations side by side. They self-register the deprecated classes on benchmark-only routes with the matching DB permission rows, since those routes are no longer part of the live app.

Case is a partial win: `solution` still needs one decode to compute the `indicators` field, so only the re-encode half of its round trip is skipped (the other four big fields — `data`, `checks`, `solution_checks`, `kpis` — skip both).

## Benchmark results (100MB per field, mean of 8 requests, fresh DB session per request)

| Endpoint | Original | Raw | Speedup |
|---|---|---|---|
| `GET /dag/<idx>/` | 4.54s | 0.95s | **4.79x** |
| `GET /instance/<idx>/data/` | 2.31s | 0.58s | **3.98x** |
| `GET /execution/<idx>/data/` | 2.43s | 0.58s | **4.19x** |
| `GET /case/<idx>/data/` | 4.95s | 2.23s | **2.22x** (partial, as expected) |

(The intermediate `DAGDetailEndpointFast` variant, tried first, showed no real improvement over the original at this payload size — see script output — which is why it was superseded by the raw-passthrough approach rather than promoted.)

Reproduce with, e.g.:
```
cd cornflow-server
python -m cornflow.tests.load.benchmark_dag_endpoint
python -m cornflow.tests.load.benchmark_instance_data_endpoint
python -m cornflow.tests.load.benchmark_execution_data_endpoint
python -m cornflow.tests.load.benchmark_case_data_endpoint
```

## Test plan
- [x] Existing test suites for `test_dags.py`, `test_instances.py`, `test_executions.py`, `test_cases.py` reused/adapted against the new routing (URL-override subclassing where a base class already existed) — 123 passed.
- [x] Full `cornflow/tests/unit` suite run both with these changes and against a stashed baseline; diffed the failure lists — identical 75 pre-existing failures in unrelated files (alarms/cli/commands/permissions/roles/schemas/signup) both times, confirming no regression from this change.
- [x] All four benchmark scripts run end-to-end at 100MB, confirming the routing swap and the deprecated-route self-registration work correctly.

Claude-Session: https://claude.ai/code/session_01UKhuvDJNSJUiXyb3zqkjAN

### DIFF
--- a/cornflow-server/cornflow/endpoints/case.py
+++ b/cornflow-server/cornflow/endpoints/case.py
@@ -5,16 +5,22 @@ These endpoints have different access url, but manage the same data entities
 """
 
 # Import from libraries
+import json
+from types import SimpleNamespace
+
 from cornflow_client.constants import INSTANCE_SCHEMA, SOLUTION_SCHEMA
 from flask import current_app
 from flask_apispec import marshal_with, use_kwargs, doc
 from flask_inflate import inflate
 import jsonpatch
+from sqlalchemy import cast
 
 
 # Import from internal modules
 from cornflow.endpoints.meta_resource import BaseMetaResource
+from cornflow.endpoints.raw_json import raw_json_object_response, restrict_to_owner
 from cornflow.models import CaseModel, ExecutionModel, DeployedWorkflow, InstanceModel
+from cornflow.shared import db
 from cornflow.shared.authentication import Auth, authenticate
 from cornflow.shared.compress import compressed
 from cornflow.shared.const import VIEWER_ROLE, PLANNER_ROLE, ADMIN_ROLE
@@ -313,7 +319,12 @@ class CaseDetailsEndpoint(BaseMetaResource):
 
 class CaseDataEndpoint(CaseDetailsEndpoint):
     """
-    Endpoint used to get the data of a given case
+    DEPRECATED: superseded by :class:`CaseDataEndpointRaw`, which is now
+    routed at this endpoint's former URL. Not routed anymore -- kept
+    around because :class:`CaseDataEndpointRaw` inherits its ``put``,
+    ``delete`` and ``patch`` from here (via :class:`CaseDetailsEndpoint`
+    and this class respectively), and for the benchmark scripts in
+    ``cornflow.tests.load`` that compare it against the optimized variant.
     """
 
     ROLES_WITH_ACCESS = [VIEWER_ROLE, PLANNER_ROLE, ADMIN_ROLE]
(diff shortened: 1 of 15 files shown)


ggsdc:
I've kept the older endpoints in case you want to compare, if you prefer I can delete older ones and keep the new optimized ones only

cornflow-server/cornflow/endpoints/case.py
```diff
@@ -313,7 +319,12 @@ def delete(self, idx):
 
 class CaseDataEndpoint(CaseDetailsEndpoint):
     """
-    Endpoint used to get the data of a given case
```
AlejandraGalan, line 316: Why dont we just delete the content of this class and replace it with the content of "CaseDataEndpointRaw" instead of changing the name? If there is no change for the user, i see no need to change the name of the classes.

cornflow-server/cornflow/tests/load/_deprecated_routes.py
```diff
@@ -0,0 +1,65 @@
...
+    :param str endpoint_name: a unique Flask endpoint name for this route
+    :param view_class: the deprecated `flask_restful.Resource` subclass
+    :param list roles: role IDs to grant GET access to (typically the
+      deprecated class's own `ROLES_WITH_ACCESS`)
+    """
+    app.add_url_rule(url_rule, view_func=view_class.as_view(endpoint_name))
+    return url_rule, endpoint_name, roles
+
+
+def grant_get_permission(url_rule, endpoint_name, roles):
```
AlejandraGalan, line 41: Why do we need this?

cornflow-server/cornflow/tests/load/benchmark_case_data_endpoint.py
```diff
@@ -0,0 +1,114 @@
+"""
```
AlejandraGalan, line 1: This code should not be merged, since it is not needed for any functionality

cornflow-server/cornflow/tests/load/benchmark_dag_endpoint.py
```diff
@@ -0,0 +1,182 @@
+"""
+Benchmark comparing two implementations of the same GET endpoint
```
AlejandraGalan, line 2: I don't think we need this

cornflow-server/cornflow/tests/load/benchmark_execution_data_endpoint.py
```diff
@@ -0,0 +1,115 @@
+"""
```
AlejandraGalan, line 1: I don't think we need this

cornflow-server/cornflow/tests/load/benchmark_instance_data_endpoint.py
```diff
@@ -0,0 +1,99 @@
+"""
```
AlejandraGalan, line 1: I don't think we need this

cornflow-server/cornflow/tests/unit/test_dags.py
```diff
@@ -141,6 +141,13 @@ class TestDagDetailEndpoint(TestExecutionsDetailEndpointMock):
     - Error handling for unauthorized access
     """
 
+    def _dag_get_url(self, idx):
```
AlejandraGalan, line 144: I don't think this is needed

cornflow-server/cornflow/tests/unit/test_cases.py
```diff
@@ -959,6 +959,13 @@ def setUp(self):
             "data",
         ]
 
+    def _data_url(self, idx):
```
AlejandraGalan, line 962: I don't think this is needed

PR #762, opened 2026-09-21
Keep soft deleted rows out of the bulk update lookup

HelenaCA:
## Problem

`BaseMetaResource.post_bulk_update` looked up the existing row with a plain query
over the fields in `self.unique`, without filtering by `deleted_at`. Resending a row
that had been soft deleted matched that row and updated it in place without clearing
its `deleted_at`: the call answered `201 Updated correctly` while the record stayed
invisible on every read, and resending it never brought it back.

## Change

The lookup now skips the deleted rows on the models that have a `deleted_at` column.
A deleted row is therefore never matched: it keeps its `deleted_at` and its data, and
the row that was sent is created as a new one, visible on the reads.

A deleted row still physically holds the id it was created with. A row sent with that
same id cannot be created next to it, so that single case is rejected with an
`InvalidUsage` (400) naming the id, instead of failing later on the primary key with
an opaque integrity error. An id held by a row that is *not* deleted is left alone and
keeps failing the way it did before.

| case | result |
| --- | --- |
| row exists and is live | updated |
| nothing matches the `unique` values | created |
| the `unique` values match a deleted row, no id sent | new row created, deleted row untouched |
| the id sent is that of a deleted row | 400 naming the id |
| the id sent is that of a live row | unchanged (primary key error) |

## Tests

Adds `cornflow/tests/unit/test_meta_resource.py`, the first tests for
`BaseMetaResource`. No endpoint of cornflow exposes the bulk update — the endpoints
that use it are the ones the cli generates (`put_bulk`) — so the resource is built on
top of a model the way `cli/tools/endpoint_tools.py` builds it, with `unique = ['id']`.
Three cases are covered: a model with soft delete (`AlarmsModel`), one without
(`ViewModel`, `EmptyBaseModel`), and the models that take the id from the payload
(`RoleModel`, `DeployedWorkflow`), which is what the generated models do.

Four of the nine tests fail against the `develop` version of the method.

`CF_ALARMS_ENDPOINT=1` is needed to run them, as for the rest of the alarms tests.

## Out of scope

`temp_el` can come out empty when the payload carries none of the fields in
`self.unique`, and `filter_by()` with no arguments then matches every row, so an
arbitrary row is overwritten instead of a new one being created. This predates the
change and is left untouched; it deserves its own issue.


cornflow-server/cornflow/endpoints/meta_resource.py
```diff
@@ -111,20 +111,53 @@ def post_bulk(self, data, trace_field="user_id"):
...
             else:
                 instance = self.data_model(el)
+                idx = getattr(instance, "id", None)
+                taken_by = None
+                if has_soft_delete and idx is not None:
+                    taken_by = self.data_model.query.filter_by(id=idx).first()
+                if taken_by is not None and taken_by.deleted_at is not None:
+                    raise InvalidUsage(
+                        f"The id {idx} belongs to a deleted record and can not be reused"
+                    )
```
marioncottard, line 160: Instead of forbidding the reuse of a deleted record (which is not user-friendly), i think it would be best to allow reusing it and set its deleted_at to None. Otherwise, how do you explain to a user that they cannot recreate a row that has been deleted?

ggsdc (approved):

ampuerin (approved):

